PR #90, opened 2016-02-25
Add parsoid to universebuild for VE and Flow

Reception123:
per miraheze/mw-config/issues/384

### DIFF
--- a/modules/parsoid/manifests/init.pp
+++ b/modules/parsoid/manifests/init.pp
@@ -65,6 +65,7 @@ class parsoid {
                 'test',
                 'tochki',
                 'torejorg',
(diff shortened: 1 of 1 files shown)


modules/parsoid/manifests/init.pp
```diff
@@ -65,6 +65,7 @@
                 'test',
                 'tochki',
                 'torejorg',
+                'universebuild'
```
Southparkfan: Missing comma at end.

Southparkfan:
https://github.com/miraheze/puppet/blob/master/modules/parsoid/templates/settings.js#L9

PR #91, opened 2016-02-25
Add universebuild for VE and flow

NDKilla:


Southparkfan:
https://github.com/miraheze/puppet/blob/master/modules/parsoid/templates/settings.js#L9

PR #107, opened 2016-03-17
Simplify ssl certs

JohnFLewis:
This is the major step to essentially rewriting how we manage SSL on a puppet-level.

This commit:
- changes Varnish over to use $custom_domains and templates instead of hard-coding all the SSL config [redundant later..!]
- adds hiera configuration for **ALL** SSL certs we currently have in MediaWiki.
- adds a class for loads all SSL certs via the ssl::cert resource (**NOT** used currently, old loading happens still).
- Converts Varnish and MediaWiki nginx config away from $custom_domains to use hiera via hiera_hash and pairing of key values using name 'property'.

After this commit is merged:
- This should be carefully and watchfully rolled out separated across cp1, cp2, mw1, mw2 for any irregularities and missing code blocks
- Clean up of $custom_domains
- Convert ssl::cert to use the autoloading except. the 'special' ones (wildcard.miraheze.org, private.reviwiki.info etc.)

What can be done after all Is good:
- SSL certs can be added in 3 lines!
- Icinga can be changed to auto-monitor every SSL cert in production + manual config for 'special cases'
- Who knows what else!


Southparkfan:
southparkfan approves

JohnFLewis:
Disabled puppet on mw* and cp* (safety)

PR #165, opened 2016-07-05
Resignation

revi:
As I have been really really really really (yes, emphasizing) busy in real life, I am completely out of date for the current state of Miraheze systems to do my duties. Also, I have other more important stuffs to do for now and I don't think I'll have time in near future, I think I'd better relinquish the bits for now.

I'll be still available as phabricator helper if you ever need me, and anyway I'm still receiving phabricator mails so if you ever need my assist ping me anyway.

Thank you for your help and support during my services.

You may want to remove my sysadmin global bits and wikicreator bits in meta.


JohnFLewis:
The syntax is all valid so accepted against the wishe of everyone unfortuntely.

Always welcome back though.

PR #201, opened 2016-08-03
Possible fix for T581

samtarling:
Possible fix for T581

Adding 301 redirect from www.thinkingliquid.org to thinkingliquid.org, guessed SSL cert locations based on pattern, filed alphabetically.

Please review this and prepare to roll back

### DIFF
--- a/modules/varnish/templates/mediawiki.conf
+++ b/modules/varnish/templates/mediawiki.conf
@@ -103,6 +103,21 @@ server {
 
 server {
 	listen 443 ssl http2;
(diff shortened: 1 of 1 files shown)


modules/varnish/templates/mediawiki.conf
```diff
@@ -103,6 +103,21 @@ server {
 
 server {
 	listen 443 ssl http2;
+	listen [::]:443 ssl http2;
+	
+	server_name www.thinkingliquid.org
```
JohnFLewis: Needs ;

PR #210, opened 2016-08-06
Create botconfig.json.erb

samtarling:


modules/phabricator/templates/botconfig.json.erb
```diff
@@ -0,0 +1,8 @@
+{
+  "server": "chat.freenode.com",
```
Reception123: I think its "irc.freenode.net" not "chat.freenode.com"

samtarling:
Fixed :+1:

PR #216, opened 2016-08-11
add certificate for aman.info.tm

Reception123:


hieradata/ssl.yaml
```diff
@@ -133,3 +133,7 @@ redirects:
     redirect: 'wikiparkinson.org'
     sslname: 'wikiparkinson.org'
     ca: 'StartSSL'
+  amaninfotm:
+    url: 'aman.info.tm'
+    sslname: 'aman.info.tm'
```
NDKilla: the "sslname" line should be removed and the entire section should be moved up (out of redirects)

Reception123:
@NDKilla done

PR #222, opened 2016-08-17
redirect Requests_for_adoption because of stupid sitenotice mistake t…

Reception123:
…hat was repeated @JohnFLewis @Southparkfan @NDKilla


Southparkfan:
Don't declare a sub twice; just put the if () { } in the other url_rewrite sub. Also, you forgot two times to add the double quotes at the end.

Southparkfan:
You still need to fix the lacking double quotes.

Reception123:
@Southparkfan Sorry, didn't notice. Fixed.

PR #263, opened 2016-10-03
enable parsoid (cybersecuritywiki) VE

ImBoPhil:


modules/parsoid/manifests/init.pp
```diff
@@ -82,6 +82,7 @@
                 'cnv',
                 'cssandjsschoolboard',
                 'cvsmb',
+                'cybersecurity",
```
NDKilla: ending " needs to be a '

PR #267, opened 2016-10-06
mediawiki::dumps implementation

NDKilla:
Puppet class and implementation to make mw2 export static dumps of select wikis.

### DIFF
--- a/modules/mediawiki/manifests/dumps.pp
+++ b/modules/mediawiki/manifests/dumps.pp
@@ -1,0 +1,16 @@
+# class: mediawiki::dumps
+#
+# Cron jobs of select wiki dumps
+class mediawiki::cron {
+	if $::hostname == 'cp1' {
+	    cron { 'update_database_lists':
+        	ensure		=> present,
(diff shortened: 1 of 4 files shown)


modules/mediawiki/manifests/dumps.pp
```diff
@@ -0,0 +1,14 @@
+# class: mediawiki::dumps
+#
+# Cron jobs of select wiki dumps
+class mediawiki::dumps {
+    cron { 'sterbalssundrystudieswiki':
+        ensure   => present,
+        command  => '/usr/bin/nice -n19 /usr/bin/php /srv/mediawiki/w/maintenance/dumpBackups.php --wiki sterbalssundrystudieswiki --logs --full > /mnt/mediawiki-static/dumps/sterbalssundrystudieswiki.xml',
+        user     => 'www-data',
+        minute   => absent,
```
JohnFLewis: these should be 0

modules/mediawiki/manifests/dumps.pp
```diff
@@ -0,0 +1,14 @@
...
+        hour     => absent,
```
JohnFLewis: 0

PR #277, opened 2016-10-17
Enable parsoid on karniarutheniawiki

LegoFan4000:


modules/parsoid/manifests/init.pp
```diff
@@ -124,6 +124,7 @@
                 'izanagi',
                 'janesskillspack',
                 'justinbieber',
+                'karniaruthenia'
```
JohnFLewis: needs a comma at the end

PR #281, opened 2016-10-22
add amaninfowiki dump

Reception123:


JohnFLewis:
Week is not a valid parameter. Even if it was, this would say 'Week two of every month' which is monthly and not what is described above.

Reception123:
@JohnFLewis SPF sent me a link but I'm not exactly sure what parameter would fit here.

NDKilla:
The current code would work but the question is do we really want to do this biweekly? also their aware that even running it bi-weekly, only the most up-to-date backup will be kept on static.mh.o?

JohnFLewis:
"The current code would work"

@NDKilla false statement!

modules/mediawiki/manifests/dumps.pp
```diff
@@ -4,11 +4,20 @@
...
         monthday => 1,
+     }
+    cron { 'Export amaninfowiki xml dump monthly':
+        ensure   => present,
+        command  => '/usr/bin/nice -n19 /usr/bin/php /srv/mediawiki/w/maintenance/dumpBackup.php --wiki amaninfowiki --logs --full > /mnt/mediawiki-static/dumps/amaninfowiki.xml',
+        user     => 'www-data',
+        minute   => '0',
+        hour     => '0',
+        month    => '*',
+        monthday => 15, 30,
```
JohnFLewis: nope!

monthday => [ '15', '30' ],

NDKilla:
shit john this is why i freaking hate puppet

NDKilla:
@JohnFLewis you can fix it in the freaking thingy >:(

JohnFLewis:
But if I fix it, people don't learn why its wrong for the future! It's not like this has to be merged right now with urgency - otherwise I would have.

Reception123:
So what should be done in this case?

Reception123:
@Southparkfan @JohnFLewis @NDKilla User is waiting for updates on this.

PR #289, opened 2016-10-30
enable parsoid (wabcwiki) VE

ImBoPhil:


ImBoPhil:
for miraheze/mw-config/pull/887

PR #292, opened 2016-11-02
add wiki.a22.dk certificate

Reception123:


Reception123:
Will e-mail the requestor and tell them to search for another SSL provider.

PR #299, opened 2016-11-11
add alwiki.net cert

Reception123:


JohnFLewis:
Once the conflict gets resolved, this can be merged. The private key is in git already.

Reception123:
fixed in #305

PR #302, opened 2016-11-12
enable oarsoid on applebranchwiki

LegoFan4000:


LegoFan4000:
Was done before #301 but forgot to create pull request

PR #307, opened 2016-11-14
add LocalisationUpdate update.php cron job T923

ImBoPhil:


modules/mediawiki/manifests/cron.pp
```diff
@@ -9,4 +9,11 @@
         minute  => '*',
         hour    => '*',
     }
+        cron { 'update.php for LocalisationUpdate':
+        ensure  => present,
+        command => '/usr/bin/php /srv/mediawiki/w/extensions/LocalisationUpdate/update.php --wiki extloadwiki',
+        user    => 'www-data',
+        minute  => '*',
```
JohnFLewis: You want to run this every minute?

modules/mediawiki/manifests/cron.pp
```diff
@@ -9,4 +9,11 @@
...
+        hour    => '24',
```
JohnFLewis: There is no 24th hour. 00:00 - 23:59.

PR #311, opened 2016-11-17
Revert "add refreshed skin composer configuration"

LegoFan4000:
Refreshed with composer has never been tested according to the MediaWiki page. This could be the cause of the problem.


JohnFLewis:
No change.

LegoFan4000:
?

JohnFLewis:
There's no change submitted. It's a blank revert commit. There is nothing to merge.

PR #312, opened 2016-11-17
Add preload to HSTS

revi:
Not sure if this will affect the custom domain, but submitting anyway. Please correct me if wrong.

Thanks for review anyway, Referenced commit d9adcf4042ac256c191ea5b4c1f77607a6bd41e0.

### DIFF
--- a/modules/mediawiki/templates/mediawiki.conf
+++ b/modules/mediawiki/templates/mediawiki.conf
@@ -16,7 +16,7 @@ server {
 
 	ssl_trusted_certificate /etc/ssl/certs/GlobalSign.crt;
 
-	add_header Strict-Transport-Security "max-age=15768000; includeSubDomains";
+	add_header Strict-Transport-Security "max-age=15768000; includeSubDomains; preload";
 
 	location /.well-known/acme-challenge/ {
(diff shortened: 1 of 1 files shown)


revi:
Ping @Southparkfan.

revi:
ping @Southparkfan

Southparkfan:
I have already discussed (along with asking questions) this a lot of times in the past and recently, but not on GitHub (and not with you actually) so I'll paste my comment here (and hopefully finally have a response from soneone).

Do we realize that if we merge this, there is no way back? If we deem it necessary to disable HSTS (which is partially possible, by setting max-age=1 and hoping most visitors will visit the wikis in the period between when we set max-age=1 and when we disable HTTPS/introduce self-signed HTTPS), that is not possible anymore. **Miraheze will always need to present a valid wildcard certificate, forever.**

Southparkfan:
@JohnFLewis says it's OK.

Will merge when the db crisis is over.

Southparkfan:
Ok, I totally forgot about this, but I think it's better to merge this when I have time to merge & deploy at the same moment.

Scheduled for tomorrow.

PR #313, opened 2016-11-19
Remove expired miraheze.wiki cert

LegoFan4000:


Reception123:
We still own the domain, the certificate needs to be renewed not removed. Please close this PR.

PR #334, opened 2016-12-16
add amaninfo images dump per req T905

Reception123:


Reception123:
@miraheze/operations this should be preferably done before 30 december so that the user can have their image dump

PR #337, opened 2016-12-19
savagewikiwiki parsoid config

LegoFan4000:


JohnFLewis:
I'll merge this but in future, old entries must be removed.

PR #348, opened 2016-12-28
Enable VE parsoid for fusewiki per T1214

DeltaQuad-DQ:


labster:
Invalid syntax.

DeltaQuad-DQ:
@labster I would appreciate it if you didn't completely close the PR just because of a syntax error - submit a review requesting changes instead and I can fix them within the same request

labster:
There's no point in doing so, as I had pushed a corrected commit immediately.

PR #350, opened 2016-12-29
Return of the Jedi

revi:
Well, I'm not a Jedi but anyway.

For mail aliases, I'm receiving address on two email addresses.

Disclaimer: I only watched Star Wars 7.


JohnFLewis:
We lack a formal policy for this, therefore common sense shall apply for me here.

This has been brought up a bit in the past (and in fact someone people have asked him to do this) therefore I shall interpret that as "nothing has changed, previous approval applies". We should formalise this though tbh.

@Reception123 has brought up no issues, therefore he is my sanity check here.

PR #351, opened 2016-12-29
add www.soundbox.wiki certificate

Reception123:


JohnFLewis:
/root/acme-tiny/ssl/soundbox.wiki.crt is signed for soundbox.wiki and www.soundbox.wiki. A new cert is not necessary

PR #352, opened 2016-12-29
renew publictestwiki.com cert

Reception123:
Done using -r


JohnFLewis:
Urm, (this isn't a LE cert)[https://github.com/miraheze/puppet/blob/master/hieradata/ssl.yaml#L16

PR #360, opened 2017-01-09
enable parsoid (fantasia) VE

ImBoPhil:


modules/parsoid/manifests/init.pp
```diff
@@ -116,6 +116,7 @@
                 'extload',
                 'ezdmf',
                 'fablabesds',
+                'fantasia'
```
JohnFLewis: missing comma

PR #369, opened 2017-02-03
renew universebuild.com cert

Reception123:


JohnFLewis:
Not LE.

revi:
I can't check if the cert is revoked or not - [crt.sh/78911361](https://crt.sh/?id=78911361) doesn't show its revocation status. (Maybe because LetsEncrypt doesn't have CRL at all) So if the cert is not revoked, please do so.

PR #371, opened 2017-02-04
attempt redirect studynotekrwiki to jayuwikiwiki

Reception123:
Not sure if this was done right (https://phabricator.miraheze.org/T1351)


labster:
Doesn't this need to be inside a server {} block?  I admit I'm not actually sure about this.

JohnFLewis:
This  could easily be handled by using the redirect set up in hiera.

Reception123:
Done in #373

PR #372, opened 2017-02-04
create changemy.org certificate (+www)

Reception123:


hieradata/ssl.yaml
```diff
@@ -160,3 +163,8 @@ redirects:
     redirect: 'soundbox.wiki'
     sslname: 'soundbox.wiki'
     ca: 'LetsEncrypt'
+ wwwchangemy:
+    url: 'www.changemy.org'
+    redirect: 'changemy.org'
+    sslname: 'changemy.org'
```
JohnFLewis: You're saying the cert for changemy.org works but I see a www.changemy.org cert as well.

PR #373, opened 2017-02-05
redirect studynotekrwiki to jayuwikiwiki

Reception123:


JohnFLewis:
Almost!

The wiki doesn't have an SSL certificate because it's a subdomain. So sslname and ca need to be changed to reflect what SSL certificate it actually uses.

PR #377, opened 2017-02-10
add bettermedia.info certificate

Reception123:


Reception123:
@NDKilla user is asking about status of cert

PR #383, opened 2017-02-24
Renew wiki.valentinaproject.org cert

Reception123:
Received email from user with new cert


JohnFLewis:
This isn't an LE cert.

I'd love to say I merged this on purpose (as a way to test our the new autoreloading I just added) but I didn't check so it was great the new system works.

Reception123:
@JohnFLewis  I know it isn't LE. I sent an email to the founder and they sent me a cert. That is what I provided

JohnFLewis:
@Reception123 okay well it doesn't match our key.

PR #385, opened 2017-02-24
Renew wiki.drones4nature.info cert

Reception123:
-r


modules/ssl/files/certificates/wiki.drones4nature.info.crt
```diff
@@ -1,8 +1,9 @@
+
```
JohnFLewis: pedantic but can we get rid of this white space

PR #387, opened 2017-02-25
Require email verification

revi:
Ref [T1487](https://phabricator.miraheze.org/T1487). The user need to verify their email to get their email. It also shows black dots if email is not verified. Enforcing email verification requires additional steps for users but it makes sure user is notified of the change on the ticket.


Reception123:
I'm not sure about this. I think this makes the feature request process more lengthy.

revi:
They're coming here to report something/request something and if they do not come back, we can never proceed. We must validate that users can get their emails so they know they have something to reply.

PR #396, opened 2017-03-11
change HSTS for 15768000 and add www.histories.wiki preload

Reception123:
not sure if we allow this max age... T1536


JohnFLewis:
This will add the header twice for every wiki not matching either two stanzas. When it matches, it'll add two conflicting headers.

More so this will never match ever as www.histories.wiki isn't a wiki.

PR #397, opened 2017-03-13
change wiki.valentinaproject.org to LE

Reception123:
per request. Site is down. Please merge this ASAP. @miraheze/operations


JohnFLewis:
What would have helped this get merged earlier is using the -r param. As the way this was done requires the key to be changed which massively delays the process (and based on the past, can lead up to a week delay).

Reception123:
-r? The cert was originally StartSSL

JohnFLewis:
-r generates a LE cert using the CSR without changing the private key or the CSR.

Reception123:
Ok. Didn't know. Will do that in the future

PR #399, opened 2017-03-15
add icmscholarswiki and scruffywiki dumps

Reception123:
Will be sent via email until we figure out a different solution.


Reception123:
Too many different commits (integrating importimages.php now). Creating separate PRs for each dump as only one has indicated that they want importImages.php

PR #400, opened 2017-03-15
add icmscholarswiki and scruffywiki dumps + alphabetize

Reception123:
Apparently the import thing isn't possible either so sticking with manual.

### DIFF
--- a/modules/mediawiki/manifests/dumps.pp
+++ b/modules/mediawiki/manifests/dumps.pp
@@ -48,7 +48,7 @@ class mediawiki::dumps {
     
     cron { 'Export scruffywiki xml dump weekly':
         ensure   => present,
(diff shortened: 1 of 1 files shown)


modules/mediawiki/manifests/dumps.pp
```diff
@@ -54,4 +45,24 @@
         month    => '*',
         monthday => ['15', '30'],
     }
+    
+    cron { 'Export scruffywiki xml dump weekly':
+        ensure   => present,
+        command  => '/usr/bin/nice -n19 /usr/bin/php /srv/mediawiki/w/maintenance/dumpBackup.php --wiki scruffywiki --logs --full > /mnt/mediawiki-static/privatedumps/scruffywiki.xml',
```
JohnFLewis: this directory will be public

PR #403, opened 2017-03-19
add yeoksa.wiki cert

Reception123:
I've checked the other Comodo cert, https://github.com/miraheze/puppet/blob/master/modules/ssl/files/certificates/oyeavdelingen.org.crt and it seems like it doesn't have an intermediate cert. As the user didn't send one for this cert either, I'm guessing it's not necessary?


JohnFLewis:
No, it needs one. That could potentially be our staple warning for the past several months.

PR #411, opened 2017-03-26
add IRC nickname option to Phab

Reception123:
why not?


Reception123:
I know this is not really important but it's been waiting for more than a week.

PR #416, opened 2017-04-07
switch 5 more StartSSL to LE

Reception123:
done using -g


JohnFLewis:
The title says 6, I count 5. Issue or mistype?

Reception123:
I don't know how I managed to type 6 instead of 5 twice :(

PR #417, opened 2017-04-10
add wiki.teessidehackspace.org.uk cert

Reception123:


Reception123:
Redo as too many conflicts

PR #420, opened 2017-04-16
move "sysadmin" to "staff"

Reception123:
sysadmins include mw-admins and all staff are currently syadmins so I think it's better than redirecting them to operations


JohnFLewis:
This was for backwards compatibility and the one account that uses this (I'm not sure what it is now?) so I'm going to reject it as its original purpose was for me and SPF only and its use was always that.

Plus to my knowledge it receives nothing for spam now.

PR #423, opened 2017-04-18
www for wikiletras.tk per req

Reception123:


hieradata/ssl.yaml
```diff
@@ -187,3 +187,8 @@ redirects:
     sslname: 'www.histories.wiki'
     ca: 'LetsEncrypt'
     hsts: 'strict'
+  wikiletras:
+    url: 'www.wikiletras.tk'
+    redirect: 'wikiletras'
```
JohnFLewis: https://wikiletras/ doesn't exist.

PR #424, opened 2017-04-18
create www cert for wikiletras.tk

Reception123:
user wants the redirect to be from wikiletras.tk to www.wikiletras.tk


JohnFLewis:
You're adding a certificate but not using it + redirecting from a registered custom domain to one that now no longer exists in ssl.yaml

Reception123:
User just needs the www, there's no need for the redirect then.

Reception123:
I'd rather ask the user to make sure, though

Reception123:
"I need www.wikiletras.tk only. Thanks. DoerNoc (talk) 20:08, 18 April 2017 (UTC)"

PR #432, opened 2017-05-07
Set Phabricator alt domain

revi:
Fixes [T1752](https://phabricator.miraheze.org/T1752). config for phab settings and apache to make server accept requests to alt-domain.


JohnFLewis:
The SSL certificate doesn't exist on misc1. It has to be added through ssl::certificate

PR #440, opened 2017-05-21
slightly increase redis maxmemory

Reception123:
See T613 comment, could reduce/fix login issues, not sure if this is the way to go though


JohnFLewis:
Memory usage on redis is 511MB. I'm not happy there is enough memory we can safely increase to give to Redis however.

I feel like this will be a merge and it'll fill up again. I'll take a look into it in a little bit more detail for now.

JohnFLewis:
~300M is unused right now. 100MB can be somewhat safely allocated but I imagine it'll fill up quickly again.

PR #447, opened 2017-06-08
add embobada.com cert

Reception123:


Reception123:
Service no longer in use. Moving to miraheze/ssl

PR #448, opened 2017-06-11
add reversible.work cert

Reception123:
Not LE, can be merged directly.


Reception123:
Nevermind, forgot that even though it's not LE, keys are bound to CSRs so therefore still need to be added. I'm guessing if the puppet-user system works, it'll be up to me to add this.

PR #453, opened 2017-06-19
Revert "add reversible.work cert"

Reception123:
Reverts miraheze/puppet#448

Meh, puppet is failing because key is missing (explained in original PR), should be reverted (as I doubt the key will be added or the puppet-users system will be in the next few hours)


Reception123:
@NDKilla @labster It is urgent that this is merged or else no changes can be made to mw-config!

PR #460, opened 2017-07-12
add "Stalled" option

Reception123:
there are a few tasks (such as upstream) where there is virtually nothing Miraheze can do. There are also other tasks that would fit into stalled, but those are the main ones.


Reception123:
(as I said on IRC) I took this variable from WMF so it should be correct

PR #462, opened 2017-07-14
add Google Maps API key per T696

Reception123:
Ops need to generate one and add it to the file first, of course.


modules/mediawiki/manifests/init.pp
```diff
@@ -134,6 +134,7 @@
     $mediawiki_secretkey  = hiera('passwords::mediawiki::secretkey')
     $recaptcha_sitekey    = hiera('passwords::recaptcha::sitekey')
     $recaptcha_secretkey  = hiera('passwords::recaptcha::secretkey')
+    $googlemaps_key  = hiera('passwords::mediawiki::googlemapskey')
```
JohnFLewis: minor but align please

PR #477, opened 2017-07-20
serve phab.miraheze.wiki .well-known reqs to mw1

Reception123:
to be able to renew the cert.
JohnLewis> I guess (no promises, ain't my area :P)


JohnFLewis:
I only just realised why is this in the mediawiki nginx config on cp* and mw*?

Reception123:
#478

PR #479, opened 2017-07-24
install HTMLDoc for PdfBook

Reception123:
There's probably some mistakes so needs review first.
@JohnFLewis

### DIFF
--- a/modules/mediawiki/manifests/init.pp
+++ b/modules/mediawiki/manifests/init.pp
@@ -174,6 +174,15 @@ class mediawiki {
         require     => Git::Clone['MediaWiki core'],
     }
     exec { 'maps_composer':
+        command     => 'curl -sS https://github.com/michaelrsweet/htmldoc.git | apt-get install htmldoc',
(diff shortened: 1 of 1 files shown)


JohnFLewis:
The name is bad (it's but maps) and why are you curling an entire git repo (won't work) and then why are you not using that git repo but instead using apt? Isn't package {} not enough?

Reception123:
@JohnFLewis Hmm, not sure what I should do instead. Is cloning the repo as a submodule a better idea?

JohnFLewis:
I don't know what you're trying to achieve. Looking at what you're trying to do above, yes a submodule

Reception123:
Ok, I'll do that. It's for https://www.mediawiki.org/wiki/Extension:PdfBook#Installation

Reception123:
@JohnFLewis HTMLDoc is not php so will a require_once work, or should it be cloned directly into the PdfBook repo?

JohnFLewis:
I've just looked at the link above and it mentions nothing about source code unless compiling. Which brings me back to my earlier question, why not just use package {}?

Reception123:
Changed. @NDKilla's PR load is back to 5 again ;)

PR #481, opened 2017-07-25
@miraheze.org alias for all staff

Reception123:
Discussed a bit in -staff with labster. I think that it is better (and more professional) if all staff have a @miraheze.org alias. It's currently a bit confusing that 3 sysadmins do have one, and 3 sysadmins don't.

### DIFF
--- a/modules/postfix/files/aliases
+++ b/modules/postfix/files/aliases
@@ -30,6 +30,12 @@ cvt: stewards,utilizator.receptie123@gmail.com,revi
 # finances
 donate: bslaabs@gmail.com,southparkfan
 
+# imbophil
+imbophil: imbophilmh@gmail.com
+
+# labster
+labster: bslaabs@gmail.com
+
 # southparkfan
 southparkfan: southparkfan223@hotmail.com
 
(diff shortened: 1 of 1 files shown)


labster:
Approved for me.  Everyone else happy getting more emails this way?

Reception123:
@labster It wouldn't be "more" emails? It's exactly the same as before only now you can get emails as username@miraheze.org

JohnFLewis:
Plus there's no change except username@ forwards to you. As far as I'm aware, no one really ever used (or wanted when asked) these except me and NDK anyway and the only one who can send miraheze.org emails is NDK. SPF dropped from an email to an alias for unknown reasons however.

Reception123:
After this is merged, I would maybe be interested in being able to send from @miraheze.org as well, but this PR itself doesn't affect that in any way.

JohnFLewis:
Then you don't want an alias like this - you want an email account. https://meta.miraheze.org/wiki/Tech:Mail

Reception123:
But I also want to receive emails on my gmail one

JohnFLewis:
Annoying setup but okay

PR #490, opened 2017-07-27
Uninstall unused phabricator applications

revi:
Ref [T1934](https://phabricator.miraheze.org/T1934). Hush.


JohnFLewis:
That task means nothing b to me but +1 for keeping my running line of slightly comedy sum up lines!

Phame was re-enabled because I think it was @labster (?) wanted to do something with it. Nothing's happened so it's fine.

PR #494, opened 2017-08-07
switch branch to REL1_29

Reception123:
[DO NOT merge unless you are upgrading]

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -3,4 +3,4 @@ mailserver: false
 arcanist: false
 mwdumps: false
 puppetmaster: false
-mediawiki::branch: 'REL1_28'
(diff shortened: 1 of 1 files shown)


JohnFLewis:
This is not the way it will be done.

This way will result in downtime. You want a phased method. Upgrade a depooled server, swap depools then repool.

Reception123:
Oh, I thought that, but wasn't sure if that was needed with test1. Well then closing this since @NDKilla can take care of the commits.

JohnFLewis:
Though this is fine if puppet is DISABLED before this is merged.

Reception123:
<JohnLewis> Reopen it. Disabling is less ugly

PR #503, opened 2017-08-31
install lilypond package

Reception123:
Needed for an extension (T2142), and could be a possible fix (or at least attempt for T1813).


JohnFLewis:
This is 100% unrelated to T1813

PR #504, opened 2017-09-01
install netpbm package

Reception123:
it says it should be for DjVu on the docs so might as well give it a try


JohnFLewis:
netpbm-procs doesn't exist and netpbm is installed as a dependency of DjVu libraries.

PR #505, opened 2017-09-30
rm non-automatic backups

Reception123:
There's no need to have automatic backups for a non-automatic process.

Until we set up a secure directory with a password/identification for individual dumps or find an automatic way to send via email these dumps are useless/unused.


Reception123:
changed.

PR #513, opened 2017-10-07
rewrite oecumene.wiki to be /$1 instead of /wiki/$1

Reception123:
Not 100% sure about the config, it should be double checked before merged.


Reception123:
@JohnFLewis would this work?

JohnFLewis:
The write happens in https://github.com/miraheze/puppet/blob/master/modules/mediawiki/files/nginx/mediawiki-includes.conf

I imagine this would make too many redirects, pot indefinitely

PR #523, opened 2017-11-11
edit backend error message

Reception123:
Put Twitter first since Twitter is updated the most, add extra message and embed Twitter feed.


Reception123:
@JohnFLewis ^ any issues with this?

JohnFLewis:
It's HTML soo

PR #524, opened 2017-11-11
Adding TME XML dump

CnocBride:
Per my request T2412. This is my first time actually changing the config itself so if there are any errors please point them out so I can rectify them :)


Reception123:
@CnocBride It's correct :)

PR #535, opened 2018-01-07
Robots Improvement

Mehran-Baghi:
Hi

Current robots.txt doesn't prevent search engines from indexing diffs or revisions. 

> spiders may try to index thousands of similar pages, overloading the webserver. [reference](https://www.mediawiki.org/wiki/Manual:Robots.txt#Prevent_indexing_of_non-article_pages)

e78c0c5 fixes that according to mediawiki's manual on robots.txt like other mediawiki farms.
[this](https://github.com/miraheze/puppet/compare/master...Mehran-Baghi:robots-improvement?expand=1#diff-294b0003d876502ab99f22e56cf95283R4) is equal to [this](https://github.com/miraheze/puppet/compare/master...Mehran-Baghi:robots-improvement?expand=1#diff-294b0003d876502ab99f22e56cf95283L3) and [this](https://github.com/miraheze/puppet/compare/master...Mehran-Baghi:robots-improvement?expand=1#diff-294b0003d876502ab99f22e56cf95283L4).

6cd3644 allows Internet Archive to <del>only</del> index raw wikitext of pages. I recommend this so in case of downtimes or lost we will have a redundant backup. [reference](https://www.mediawiki.org/wiki/Manual:Robots.txt#Allow_indexing_of_raw_pages_by_the_Internet_Archiver)

### DIFF
--- a/modules/mediawiki/files/robots.txt
+++ b/modules/mediawiki/files/robots.txt
@@ -7,3 +7,7 @@ Disallow: /wiki/Special:
 # Throttle YandexBot
 User-Agent: YandexBot
 Crawl-Delay: 2.5
+
+# Allow the Internet Archiver to index action=raw and thereby store the raw wikitext of pages
+User-agent: ia_archiver
+Allow: /*&action=raw
--- a/modules/mediawiki/files/robots.txt
+++ b/modules/mediawiki/files/robots.txt
@@ -1,7 +1,7 @@
 # Disallow API and special pages
 User-Agent: *
-Disallow: /w/api.php
-Disallow: /w/index.php?title=Special:
+Allow: /w/load.php?
+Disallow: /w/
 Disallow: /wiki/Special:
 
 # Throttle YandexBot


Mehran-Baghi:
Ok, I didn't know about WikiTeam.
What about e78c0c5 ?

modules/mediawiki/files/robots.txt
```diff
@@ -1,9 +1,13 @@
...
+Disallow: /w/
 Disallow: /wiki/Special:
 
 # Throttle YandexBot
 User-Agent: YandexBot
 Crawl-Delay: 2.5
+
+# Allow the Internet Archiver to index action=raw and thereby store the raw wikitext of pages
+User-agent: ia_archiver
+Allow: /*&action=raw
```
Southparkfan: Is &action=raw ever used by IA?

Mehran-Baghi: Yes, under Miraheze's current robots.txt directives, IA is able to archive raw pages. I just saved the main page with &action=raw. [Link](https://web.archive.org/web/20180224181708/https://meta.miraheze.org/wiki/Miraheze?action=raw).

Mehran-Baghi:
@revi @Southparkfan 

IA's Wayback Machine archives html pages (as usual) and wikiteam backs xmls up (using [these scripts](https://github.com/WikiTeam/wikiteam)).
Wikiteam's latest backup is for [September 2016](https://archive.org/details/wikifarm-miraheze.org-20160930) at 685 wikis which is quite old. However Wayback Machine might have more versions of wikis. (for example look at [this link](https://web.archive.org/web/*/https://meta.miraheze.org/wiki/Miraheze).)

In case of a data loss, if there won't be a xml dump, there might be a html page at Wayback Machine. However raw text is also quite useful. By allowing Wayback Machine to pass the `&action=raw` to the pages and archiving wikitexts, we will be able to use them on new wikis.

@Southparkfan 

Note that editing the robots.txt to `Allow: /*&action=raw` isn't enough. As you can see in [this link](https://web.archive.org/web/*/https://meta.miraheze.org/wiki/Miraheze?action=raw) there are no snapshots in the past for wikitexts. The reason is IA still needs to find the links to the raw pages to crawl. According to [this](https://www.mediawiki.org/wiki/Manual:Internet_Archive#Getting_data_from_your_wiki_into_the_Internet_Archiver) we have to either list them in an `AllPages` or have a link on side bar using `**{{fullurl:{{FULLPAGENAMEE}}|action=raw}}|View raw wikitext`

JohnFLewis:
Per the above, probably something up for discussion? I'm unsure but with no movement this is a dead PR as far as I am concerned currently. The first change seems fine from reading so that can be resubmitted separately.

PR #542, opened 2018-01-10
add /etc/puppet/ssl to basemodulepath

paladox:


JohnFLewis:
This isn’t a module though. It’s a list of files only.

PR #597, opened 2018-02-05
Remove unecessary double comma

sau226:
In a config file there is a double comma where there is only meant to be one. This PR fixes it (and maybe certain errors related to it).


Reception123:
Thanks for pointing this out / fixing.

PR #608, opened 2018-02-28
Update puppet-master.conf.erb

paladox:


paladox:
Will re open when i fix an issue.

PR #624, opened 2018-03-29
Purge icinga resources where it is not exported

paladox:


paladox:
Untested but should work.

cc @Reception123

paladox:
https://github.com/wikimedia/puppet/search?utf8=✓&q=collect+exported+resources&type=

PR #625, opened 2018-03-30
Piwik rename to Matomo

Reception123:
Since Piwik has been renamed to Matomo, I think we should also change the name to reflect this change.

@JohnFLewis If I rename the database as well, this shouldn't break anything, right?


JohnFLewis:
Will deal with later.

Southparkfan:
Since piwik/matomo is not a critical service imho I would not decline this pull request (it is impossible to avoid downtime unless you allow analytics to be served from both subdomains at the *same time*), but keep in mind you cannot perform other migrations in the same way.

Besides, what about analytics.miraheze.org? :)

Southparkfan:
@JohnFLewis I'm fine with this PR merge but I would like a response to my proposal (see last question). Is analytics too vague?

JohnFLewis:
My personal proposal is get rid of this eventually anyway because it doesn't seem to suit the needs we need moving forward of useful analytics for everyone. And if it can, it doesn't now so someone has to take control of this and deal with it.

PR #639, opened 2018-04-02
Up the puppet runs to 30mins instead of 10 mins

paladox:


Reception123:
-1 don't think the advantages outweigh the disadvantages here tbh

PR #641, opened 2018-04-03
Update mediawiki puppet class to support php 7.0 if on stretch

paladox:
This will use php 7.0 on the mediawiki class.

This will also support any other changes that were done in stretch.


modules/mediawiki/manifests/init.pp
```diff
@@ -208,8 +217,15 @@
...
     }
 
-    icinga::service { 'php5-fpm':
-        description   => 'php5-fpm',
-        check_command => 'check_nrpe_1arg!check_php_fpm',
-    }
+    if os_version('debian >= stretch') {
+        icinga::service { 'php7.0-fpm':
+            description   => 'php7.0-fpm',
+            check_command => 'check_nrpe_1arg!check_php_fpm',
```
Southparkfan: Please be aware this nrpe command checks for 'php5-fpm' processes.

paladox: Yep :).

I am going to rename some things so we can tell apart php5-fpm and php7.0-fpm :).

PR #652, opened 2018-04-04
If monitor is false in apache::site makes sure the service is removed

paladox:


JohnFLewis:
What about nginx::site? Also what if we have two web servers but one shouldn’t be monitored? :)

JohnFLewis:
I mean, of course we’d only want to remove it if it’s defined. Then again the previous change would only remove it if it existed :)

paladox:
@JohnFLewis Yep, my second patch fixes it :)

JohnFLewis:
Nope, it does the same thing. You’re removing it if it exists in the first and now your only saying remove it if it exists.

paladox:
@JohnFLewis Yep. Since monitor = true turns both http and https on, so when we switch it off it can remove both if they exist.

JohnFLewis:
But it removes all Icinga checks for HTTP on the server. My case described above still applies. What if we need the checks but not for one service. All other checks are removed when needed

paladox:
oh i see. Not sure then. @JohnFLewis would you know how we could do this then? :)

JohnFLewis:
Besides using $monitor = false in the first place, no.

PR #657, opened 2018-04-04
Back to mdb 10.2

MacFan4000:


paladox:
Not needed as the if check above does it.

MacFan4000:
Ignore this comment

PR #659, opened 2018-04-05
Use wildcard.miraheze.org for postgresql ssl cert

paladox:


JohnFLewis:
use include ssl::wildcard please

PR #666, opened 2018-04-05
Change !defined to defined

paladox:
This fixes an issue where it try to add the file again leading to puppet errors as you carn't do the same thing twice.


JohnFLewis:
This adds it ONLY IF it’s already added

PR #668, opened 2018-04-05
Increase puppet runs to 15 mins

paladox:


Reception123:
My last opinion still stands, I don't think changing from 10 to any number has enough benefits. I think keeping it at 10 is the best.

PR #685, opened 2018-04-10
[Do not merge] Add new mathoid module

paladox:
This module will install mathoid (requires a manual npm install after the puppet run then puppet should be re run)


modules/mathoid/manifests/init.pp
```diff
@@ -0,0 +1,112 @@
...
+    file { '/etc/systemd/system/restbase.service':
+        ensure => present,
+        source => 'puppet:///modules/mathoid/restbase.systemd',
+        notify => Exec['mathoid reload systemd'],
+    }
+
+    service { 'restbase':
+        ensure     => 'running',
+        enable     => true,
+        provider   => 'debian',
```
Southparkfan: You're sure this one shouldn't be systemd?

modules/mathoid/manifests/init.pp
```diff
@@ -0,0 +1,112 @@
...
+    file { '/etc/systemd/system/mathoid.service':
+        ensure => present,
+        source => 'puppet:///modules/mathoid/mathoid.systemd',
+        notify => Exec['mathoid reload systemd'],
+    }
+
+    service { 'mathoid':
+        ensure     => 'running',
+        enable     => true,
+        provider   => 'debian',
```
Southparkfan: You're sure this one shouldn't be systemd?

PR #688, opened 2018-04-10
Increase timeout to 600 for git::clone

paladox:


JohnFLewis:
10 minutes? I’d be concerned if something takes 10 minutes to clone, surely?

PR #701, opened 2018-04-11
Add Paladox to ops

MacFan4000:


modules/icinga/files/contactgroups.cfg
```diff
@@ -1,7 +1,7 @@
 define contactgroup{
 	contactgroup_name	ops
 	alias			Operations
-	members			southparkfan,irc,ndkilla,reception,johnflewis
+	members			southparkfan,irc,ndkilla,reception,johnflewis, paladox
```
paladox: Need to remove a space please :)

PR #708, opened 2018-04-11
Add support for php 7.2+ in icinga and phabricator class

paladox:
Phabricator requires php5 or php 7.1+.

Php 7.0 is not supported.


JohnFLewis:
A skim says it looks good

Southparkfan:
Do we seriously want to duplicate this? Can't we move the stanzas to a new class and include that class in all modules?

paladox:
@Southparkfan we could yes, i guess that would be nice :).

PR #731, opened 2018-04-26
Add me to mw-admins

MacFan4000:


JohnFLewis:
Not 4096 keys.

modules/users/data/data.yaml
```diff
@@ -7,7 +7,7 @@ groups:
   mediawiki-admins:
     gid: 2001
     description: sudo on MediaWiki servers
-    members: [revi]
+    members: [revi, macfan]
```
JohnFLewis: you're shell name is macfan4000 below

PR #741, opened 2018-05-09
Use testing branch for mw-config for testing

MacFan4000:


paladox:
https://github.com/miraheze/puppet/blob/master/modules/mediawiki/manifests/init.pp#L46

MacFan4000:
How would it be done then?

paladox:
You would need to add a new var to https://github.com/miraheze/puppet/blob/master/modules/mediawiki/manifests/init.pp#L3 named differentily otherwise it will conflict then do mediawiki::<new var>: <branch>

MacFan4000:
Mediawiki has that as well

MacFan4000:
Now I have created a REL1_31 branch on mw-config

PR #752, opened 2018-05-20
Add list.miraheze.org to postfix

paladox:


modules/postfix/files/main.cf
```diff
@@ -1,6 +1,6 @@
 myhostname = mail.miraheze.org
 myorigin = /etc/mailname
-mydestination = mail.miraheze.org, miraheze.org, localhost
+mydestination = mail.miraheze.org, list.miraheze.org, miraheze.org, localhost
```
JohnFLewis: list.miraheze.org isn’t a destination address.

paladox: Thanks, done.

PR #758, opened 2018-06-18
Don't claim a task when declining it

MacFan4000:


JohnFLewis:
Declining is like resolving, you've made a decision therefore it is your task.

PR #775, opened 2018-07-13
optimising php-fpm

paladox:


JohnFLewis:
5 minute request time out? What request is expected to take 5 minutes

PR #780, opened 2018-07-23
automatically update Miraheze extensions

Reception123:
As you know, I am not too good with these things, so please double check and tell me what is wrong and what needs to be changed.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -2,6 +2,7 @@ jobrunner: false
 mailserver: false
 arcanist: false
 mwdumps: false
+extensionupdate: false
 puppetmaster: false
 mediawiki::branch: 'REL1_31'
 acme: false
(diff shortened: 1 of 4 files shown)


JohnFLewis:
1) There’s no auth involved here so GH will reject push attempts.
2) This is trying to automate something for no reason.
3) It’s messy.
4) I have several times made commits to extensions and told people “Don’t deploy this until x is done” or maybe I’m doing a large amount of work on an (ManageWiki+CreateWiki in the past) which requires things like Scheme changes. This doesn’t take that into account so would automatically deploy broken code.

It’s risk for no reward.

Reception123:
This PR also includes changing @MirahezeSSLBot to MirahezeGitHubBot since it will now have multiple purposes

Reception123:
ok

Reception123:
@JohnFLewis Is there a change now that can't be deployed (so I know if I can go ahead with the manual update)?

PR #781, opened 2018-07-23
Import default.vcl

paladox:


Reception123:
Why are we changing the current file?

Southparkfan:
Don't merge this

paladox:
@Southparkfan i doin't plan too.

Though we should improve this template to try and increase cache hits (and cache longer)

PR #798, opened 2018-08-02
Some changes to ssh key

paladox:


JohnFLewis:
Why www-data? Doesn’t nagiosre need this?

PR #799, opened 2018-08-04
Add miraheze specific configs for discord hook

paladox:


modules/mediawiki/templates/PrivateSettings.php
```diff
@@ -24,3 +24,7 @@
 
 // Matomo Token
 $wgMatomoAnalyticsTokenAuth = "<%= @matomotoken %>";
+
```
Reception123: Let's have a // Extension:DiscordNotifications hooks 

here

modules/mediawiki/templates/PrivateSettings.php
```diff
@@ -24,3 +24,7 @@
...
+$wmgMetaMirahezeDiscordHook = "<%= @meta_discord_hook %>";
+
+$wmgWikiMirahezeDiscordHook = <%= @wiki_discord_hook %>;
```
Reception123: wouldn't there be "" as for the meta_discord_hook?

JohnFLewis:
Why are you adding a private wiki settings file for LocalWiki? Don’t. Please don’t.

One is enough.

paladox:
Because the hook url has to stay private and PS.php is loaded far too early.

JohnFLewis:
No?

$wmg as an array, load the array variable.

You are not making another private file for no reason.

modules/mediawiki/templates/PrivateSettings.php
paladox: done

PR #808, opened 2018-08-14
redesign varnish error pages

MacFan4000:
I know that the spacing is probably off. Draft: https://test1.miraheze.org/503.html


MacFan4000:
@paladox please review. I think this may be ready.

PR #811, opened 2018-08-17
Set slow-query-log to 0

paladox:
It keeps using too much storage


JohnFLewis:
should be 0 not removed

PR #816, opened 2018-08-29
Fortnightly backup for schattenvonskeloswiki

shmith:
Per request T3527

(The XML dump process has changed, am I even doing this right?)


modules/mediawiki/data/xml_dump.yaml
```diff
@@ -11,6 +11,7 @@ nenawikiwiki: 'weekly'
 nissanecuwiki: 'fortnight'
 nonbinarywiki: 'monthly'
 renaissancewiki: 'monthly'
+schattenvonskelos: 'fortnight'
```
paladox: I think this should have wiki at the end of the name.

MacFan4000: its already been reverted @paladox

PR #823, opened 2018-09-02
Add phabricator to varnish

paladox:


JohnFLewis:
We’re not putting phab behind varnish.

PR #827, opened 2018-09-05
Retry ssl-acme up to 5 times

paladox:
This is to try and reduce race condition when alot of ssl certs need renewing thus all the checks are scheduled.


JohnFLewis:
It looks okay? Deploy if you want

PR #858, opened 2018-09-29
Add script-src 'self' to varnish

paladox:


JohnFLewis:
self is already in default

PR #895, opened 2018-11-19
Update conduct@ and cocc@

revi:
So that all mails go to mailing list.


MacFan4000:
@revi this would mean that you can remove your alias?

MacFan4000:
@revi , also I see you updated the ACL project on phab, but not the other project

MacFan4000:
@revi https://phabricator.miraheze.org/project/view/34/

MacFan4000:
Though I guess only phab admins can do anything with it. (Commenting here as I dont have IRC access at the moment)

revi:
For the completeness

1. I will do it once I clear it from my various email point
2. that project does not matter when it comes to ACL purpose: that project can be joined by anyone. As far as I know, the project to control the ACL to the tasks in the space is the one I modified.

MacFan4000:
> 2. that project does not matter when it comes to ACL purpose: that project can be joined by anyone. As far as I know, the project to control the ACL to the tasks in the space is the one I modified.
It cant be joined by anyone. The edit and join policya are restricted to phab admins (just saying)

revi:
Uh conductwiki. I forgot about that but void dealt with that. so Thanks void!

PR #898, opened 2018-11-22
Redirect /.well-known/acme-challenge/ to mw1.miraheze.org

paladox:


JohnFLewis:
Also https://github.com/miraheze/puppet/blob/ec9ecaa70e873229e2bc44f292bd36c1c090c30b/modules/varnish/templates/default.vcl#L187

PR #909, opened 2018-11-27
Update icinga mail addresses

MacFan4000:
Change non mirages staff emails to miraheze ones


JohnFLewis:
Some people may prefer to use their personal emails for this.

MacFan4000:
not really a concern, reception and ndkilla had notifications disabled, spf can’t recive mh email on hotmail, and I am fine with my notifications going to macfan@. everybody else already was using their my address.

PR #943, opened 2018-12-21
icinga: use correct http_vhost for phab and expect 200, not 500

mutante:
"This Phabricator install is configured as "https://phabricator.miraheze.org/",
but you are using the domain name "phab.miraheze.wiki" to access a page
which is trying to set a cookie..."

Do not expect/accept 500 as normal, that would be SNAFU.

Expect a 200 OK when using the actually configured http_vhost.

### DIFF
--- a/modules/phabricator/manifests/init.pp
+++ b/modules/phabricator/manifests/init.pp
@@ -140,9 +140,9 @@ class phabricator {
     monitoring::services { 'phab.miraheze.wiki HTTPS':
         check_command => 'check_http',
         vars          => {
-            http_expect => 'HTTP/1.1 500 Internal Server Error',
+            http_expect => 'HTTP/1.1 200 OK',
             http_ssl    => true,
-            http_vhost  => 'phab.miraheze.wiki',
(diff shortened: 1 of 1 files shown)


paladox:
This duplicate https://github.com/miraheze/puppet/blob/master/modules/phabricator/manifests/init.pp#L149 :)

JohnFLewis:
On a point of principle, it's 100% correct (expecting a failure and calling it good is weird to say) but on a technical point it's changing the definition of the monitoring completely. This may require a different approach entirely but it should be done.

mutante:
So which is the correct URL of Phabricator, phabricator.miraheze.org or phab.miraheze.wiki?

If it is miraheze.org then why not remove the monitoring for the wrong or outdated name?

If it is miraheze.wiki then why use a different domain for one random service while it's the service of the Miraheze Org?

mutante:
Oooh.. this is the equivalent to wmfusercontent.org , now i understand why you have this setup. Yea, then nevermind my pull request.

PR #958, opened 2019-01-03
NEEDS OPS OPINIONS promote macfan to admin in icingaweb2

MacFan4000:
Per paladox message on IRC this will need OP opinions @JohnFLewis @paladox @NDKilla @Reception123 @Southparkfan

Also note that I’m familiar with icingaweb2 since I currently run an instance of it

AFAIK the only reason why mw-admins never had accounts before was because it required a shell account on misc1


MacFan4000:
@miraheze/operations Opinions?

JohnFLewis:
> AFAIK the only reason why mw-admins never had accounts before was because it required a shell account on misc1

Nope. Only reason is they never needed to do anything and still don't AFAIK?

PR #960, opened 2019-01-04
make host monitoring work also on non-containers again

mutante:
In #951 support for openvz containers was added where you have to use a custom fact to get the virtual IP address. This removed support for regular non-containers though where the regular ipaddress fact should be used. Make the module flexible to support both.


paladox:
Thanks!

PR #977, opened 2019-01-17
Fix installing luasandbox

MacFan4000:
1. It’s only needed for MediaWiki
2. As far as I can tell the reason it isn’t working is because the INI file isn’t present

I can verify that luasandbox does indeed work with PHP 7.3 so that shouldn’t be an issue

### DIFF
--- a/modules/mediawiki/manifests/packages.pp
+++ b/modules/mediawiki/manifests/packages.pp
@@ -13,6 +13,7 @@ class mediawiki::packages {
         'libav-tools',
         'libvips-tools',
         'lilypond',
+        'php-luasandbox',
         'poppler-utils',
         'python-pip',
         'netpbm',
--- a/modules/php/manifests/php_fpm.pp
+++ b/modules/php/manifests/php_fpm.pp
@@ -85,7 +85,7 @@ class php::php_fpm(
             priority => 10,
     }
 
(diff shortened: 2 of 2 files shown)


paladox:
This wouldn't fix it. Your simply just moving the package to a seperate class.

MacFan4000:
@paladox looking at your commits I can see that you had moved it to require_package to avoid creating the INI file

MacFan4000:
Anyhow it’s clearly not being installed correctly, as from what I can tell, the ext isn’t being loaded in php7.3

paladox:
Uh, no. php module is removing the ini file. So if php is not maintaining the ini file it removes it.

MacFan4000:
@paladox I refer to 
https://github.com/miraheze/puppet/commit/f6a1c70d6ac7432161b887f17f5c1aefcde6644c#diff-b7db7294eb4ad675535964a347afe517

MacFan4000:
It certainly does work, because I have it working on my server which is using php 7.3

paladox:
How did you test?

MacFan4000:
1. I used the phpinfohook in a php file to see if it was being loaded, 2 I checked that I was seeing LuaSandbox in Special:Version on-wiki

Also see the following comment
https://github.com/miraheze/puppet/commit/f6a1c70d6ac7432161b887f17f5c1aefcde6644c#diff-b7db7294eb4ad675535964a347afe517

PR #1005, opened 2019-05-26
move The Pioneer to stewards

The-Voidwalker:
per https://meta.miraheze.org/w/index.php?oldid=73946#.E9.96.8B.E6.8B.93.E8.80.85.27s_Request_for_Stewardship


paladox:
Per https://meta.miraheze.org/w/index.php?title=Requests_for_Stewardship&type=revision&diff=73947&oldid=73946

PR #1019, opened 2019-06-05
Add a check to check for 1% of the disk

paladox:
DB4 has been running out of space lately due to the build up of bin logs files. But because it's gone passed 10% left, we are not notified.

Adding another check for 1% left will make sure we will be able to respond quickly before db4 goes down.

Bug: T4444


JohnFLewis:
Why do we need another disk space check? What is wrong with the current disk space check?

And if you say "10% is too high", fix it.

PR #1091, opened 2019-09-27
Reduce timeout to 150 for php-fpm and 140 for nginx

paladox:


paladox:
This should help with cases where a request takes too long.

PR #1143, opened 2019-11-10
Add void and owen to staff@

paladox:


JohnFLewis:
Staff is technical team (and labster). This isn’t changing, Owen and avoid aren’t staff.

PR #1160, opened 2019-12-11
vanish: Remove data schemas from CSP

paladox:
It's not recommended according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src


Southparkfan:
Status?

PR #1167, opened 2019-12-24
Update with new role titles

RhinosF1:
T5016

Ensure full consensus and check any config for maniphest

### DIFF
--- a/modules/postfix/files/aliases
+++ b/modules/postfix/files/aliases
@@ -18,13 +18,13 @@ root: /dev/null
 
 # redirect for old names
 operations: sre
-staff: sysadmins
+staff: tech
 
 # staff aliases
-wiki-abuse: sysadmins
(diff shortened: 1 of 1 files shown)


modules/postfix/files/aliases
```diff
@@ -16,22 +16,23 @@ void-recipient: /dev/null
 # root spam
 root: /dev/null
 
-# operations aliases
-ops: operations
+# redirect for old names
+operations: sre
+staff: sysadmins
```
paladox: I doint really like this but if John supports it then I won’t block this.

How about tech@? Nicer name.

RhinosF1: Tech@ sounds great and I did say I’m open to nicer names!!

RhinosF1:
Please squash on merge

PR #1172, opened 2019-12-30
Update donate@

RhinosF1:
Now handled by Owen.

Will contact Owen & Labster to notify

### DIFF
--- a/modules/postfix/files/aliases
+++ b/modules/postfix/files/aliases
@@ -47,7 +47,7 @@ rhinosf1: rhinosf1@gmail.com
 campaign: sre
 
 # board
-board: owen,southparkfan,void
+board: owen,southparkfan,void,robla
 
 # phabricator incoming emails - must be configued with Maniphest
 security: sre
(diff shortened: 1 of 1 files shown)


labster:
LGTM.  I closed the Paypal account, so there's no need for me to have access to this email now.  You should be able to reopen at the same address; let me know if there's a problem.

modules/postfix/files/aliases
```diff
@@ -47,7 +47,7 @@ rhinosf1: rhinosf1@gmail.com
 campaign: sre
 
 # board
-board: owen,southparkfan,void
+board: owen,southparkfan,void,robla
```
paladox: Seems wrong as Reception added this?

RhinosF1: I rebased it, not sure what the diff is doing

RhinosF1:
13:01:58 <RhinosF1> Leave to SPF|Cloud pls

RhinosF1:
@Southparkfan

PR #1210, opened 2020-02-12
Create cp6.yaml

sevillafunciona:


Reception123:
Created by an LTA (multiple accounts abuse)

PR #1229, opened 2020-02-14
mediawiki: Switch mw1 to jobrunner1 for letsencrypt and services

paladox:


JohnFLewis:
jobrunner1 not mw4 please

PR #1309, opened 2020-03-30
remove some old infra config

Reception123:
DO NOT MERGE unless you have reviewed and are sure it will not break anything

@paladox


modules/monitoring/files/groups.conf
```diff
@@ -40,10 +40,6 @@ object HostGroup "phabricator" {
   assign where host.name in [ "phab1" ]
 }
 
-object HostGroup "redis" {
```
paladox: You can keep this but change it rdb[12]

PR #1344, opened 2020-04-22
up sleep time to 1

RhinosF1:
Bot keeps flooding out and ideally we shouldn’t be sending more than 1 message every 2 seconds. Try upping to 1s sleep as I don’t think we’ll have any issues with that.


Reception123:
LGTM, maybe it will fix our issues. @paladox ?

PR #1379, opened 2020-05-19
add yandex API key setting for translate extension

Reception123:


modules/mediawiki/manifests/init.pp
```diff
@@ -97,6 +97,7 @@
     $recaptcha_secretkey  = lookup('passwords::recaptcha::secretkey')
     $googlemaps_key       = lookup('passwords::mediawiki::googlemapskey')
     $matomotoken          = lookup('passwords::mediawiki::matomotoken')
+    $googlemaps_key       = lookup('passwords::mediawiki::yandextranslationkey')
```
paladox, line 100: Should be yandextranslation_key

Reception123: Why is it not for google maps? https://github.com/miraheze/puppet/blob/9651400d4f605bfa730e1a445f36003855bcc135/modules/mediawiki/templates/PrivateSettings.php#L9

paladox: huh? There is already a key called 'googlemaps_key', this is a duplicate, not to mention the name is not correct for the what it'll be doing.

modules/mediawiki/templates/PrivateSettings.php
```diff
@@ -25,6 +25,9 @@
 // Matomo Token
 $wgMatomoAnalyticsTokenAuth = "<%= @matomotoken %>";
 
+// Translate Yandex API key
+$wmgYandexTranslationKey = > "<%= @yandextranslation_key @>";
```
paladox, line 29: Should be ```$wmgYandexTranslationKey = "<%= @yandextranslation_key %>";```

modules/mediawiki/templates/PrivateSettings.php
```diff
@@ -25,6 +25,9 @@
 // Matomo Token
 $wgMatomoAnalyticsTokenAuth = "<%= @matomotoken %>";
 
+// Translate Yandex API key
+$wmgYandexTranslationKey = "<%= @yandextranslation_key @>";
```
paladox, line 29: Should be ```$wmgYandexTranslationKey = "<%= @yandextranslation_key %>";```

PR #1386, opened 2020-05-29
monitoring: Increase ping critical/warning levels

paladox:
Set these vars:

```
	vars.ping_wrta = 500
 	vars.ping_wpl = 20
 	vars.ping_crta = 2000
 	vars.ping_cpl = 100
```

This means we up when we alert to 500ms+ (warning) and 2000ms (critical).

Packet loss is 20% (warning) and 100% (critical).

Defaults are:

```
	vars.ping_wrta = 100
	vars.ping_wpl = 5
	vars.ping_crta = 200
	vars.ping_cpl = 15
```

### DIFF
--- a/modules/monitoring/files/services.conf
+++ b/modules/monitoring/files/services.conf
@@ -28,6 +28,11 @@ apply Service "ping4" {
 
   check_command = "ping4"
 
+	vars.ping_wrta = 500
+	vars.ping_wpl = 20
+	vars.ping_crta = 2000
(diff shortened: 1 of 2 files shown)


JohnFLewis:
From an administrative point, I decline this change. 499ms is not an acceptable ping level for most of our servers.

RhinosF1:
The highest current value is 251ms and based on Discord none of our users in Asia have any unusual issues.

PR #1388, opened 2020-06-03
Add Reddit and Dropbox to CSP whitelist T5685

Amanda-Catherine:


Amanda-Catherine:
@Southparkfan review requested/needed

Southparkfan:
Fine with me

Amanda-Catherine:
I just copied and pasted what was given on the Phab task.

RhinosF1:
I guess we should ask then

Elaeagnifolia:
Re 59&59: As a Reddit user, I would recommend keeping these lines, as the easily retrievable/exposed URLs for uploaded Reddit videos and images use v.redd.it and i.redd.it respectively. I'm actually not sure, outside of rendering post image thumbnails, if Reddit utilizes redditmedia for serving normal videos/images (at least not visibly). Even if they still do, there's isn't an intuitive process in place to figuring out what the redditmedia URL would be.

See this image as an example:
https://b.thumbs.redditmedia.com/_OjMi-urwiRaIDc6hJ79Lk0ChrdCGLggUCVEXyEGo1o.jpg
https://i.redd.it/ksn1lcio0u051.jpg

RhinosF1:
That makes sense, if it's in use as an active url, makes sense as it's not just forwarding.

Southparkfan:
Just wanted to say I have no objections.

Reception123:
Per SPF this is good to go

paladox:
well *.x.org would only be like test.x.org it won't be test.test.x.org.

Reception123:
@paladox so is there anything we can do about that?

paladox:
@Southparkfan says it's fine, merging.

PR #1402, opened 2020-06-11
add wildcards for googleusercontent and imgbox T5614

Amanda-Catherine:


Amanda-Catherine:
@Southparkfan @paladox as SPF has no objections, can this and #1388 be merged?

Reception123:
Root domains do not work with the current system

PR #1432, opened 2020-07-07
monitoring: Add new check_mysql-replication check

paladox:


modules/mariadb/manifests/config.pp
```diff
@@ -110,4 +110,13 @@
             mysql_ssl      => true,
         },
     }
+
+    if $server_role == 'slave' {
+        monitoring::services { 'Check Mysql Replication':
```
Southparkfan, line 115: Check MariaDB Replication

PR #1455, opened 2020-07-18
Ircrcbot: change lineRate to 0.5

paladox:


RhinosF1:
This won't help at all.

PR #1463, opened 2020-07-31
check_reverse_dns: handle an exception

RhinosF1:
should make https://icinga.miraheze.org/monitoring/event/show?type=notify&id=134112 more clear, a bit of a proof of concept without tests but it works in my sleepy head.

### DIFF
--- a/modules/monitoring/files/check_reverse_dns.py
+++ b/modules/monitoring/files/check_reverse_dns.py
@@ -2,7 +2,7 @@
 """
 Script to check if reverse DNS entry for hostname matches given regex.
 
-Version: 0.1.1 (2020-07-29)
+Version: 0.1.2 (2020-07-31)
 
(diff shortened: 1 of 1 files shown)


modules/monitoring/files/check_reverse_dns.py
```diff
@@ -76,8 +76,11 @@ def main():
         """Execute functions"""
 
         args = get_args()
-
-        rdns_hostname = get_reverse_dnshostname(args.hostname)
+        try:
+                rdns_hostname = get_reverse_dnshostname(args.hostname)
+        except NoNameservers:
+                print("rDNS CRITICAL - {} All nameservers failed to answer the query.".format(args.hostname))
+
```
JohnFLewis, line 83: no exit?

RhinosF1: Fixed

PR #1466, opened 2020-08-03
403 clients with no UserAgent

RhinosF1:


Reception123:
Merging as a temporary measure due to high load. should be further reviewed by SRE

RhinosF1:
> Merging as a temporary measure due to high load. should be further reviewed by SRE

This is only intending to block the UA -

PR #1478, opened 2020-08-06
check_reverse_dns: fix NoNameservers handling

RhinosF1:


Southparkfan:
Version should be changed first.

PR #1489, opened 2020-09-03
add vimeo to CSP per T6129

Reception123:
second-largest video sharing service (after YouTube)


Reception123:
Needs another SRE to sign off first

PR #1494, opened 2020-09-18
varnish: Change rate limit rules

paladox:
* Increase rate limiting to 30r/s.
* Increase bursts to 20.
* Use nodelay (reference https://www.nginx.com/blog/rate-limiting-nginx/#Queueing-with-No-Delay).
* Introduces a static rate limiter allowing up to 300 requests a second with a burst of up to 200.


RhinosF1 (commented):
Concept and idea seems good. No clue about nginx syntax.

PR #1505, opened 2020-10-06
varnish: Increase rate limit for static to 500/1s

paladox:
The reason is static can handle much more without issues. So 120 over 10s was little especially for static which can handle even more than 500 but I think 500 is a good in between number in the meantime till I get @Southparkfan thoughts.


RhinosF1 (approved):
Thanks!

JohnFLewis:
No progress for 3 months?

paladox:
> No progress for 3 months?

needs you or SPF to sign off I guess.

JohnFLewis:
It's fine to me.

PR #1511, opened 2020-11-05
Temp comment manageInactiveWikis (due to T6403)

Universal-Omega:
I apologize if this is not the right action to take but it shouldn't be on if it's malfunctioning anyway. I am not all that familiar with puppet either, so I may not have done it correctly either.


Reception123:
I've [already absented](https://github.com/miraheze/puppet/commit/09e9000da36e3485ee0ca59c55f5472b34615c4b) the cron so it should already no longer be running.

PR #1567, opened 2020-12-22
Remove nagiosre user as not needed

paladox:


JohnFLewis:
Is this still needed?

PR #1593, opened 2021-01-08
Remove $wmgMapsGMaps3ApiKey from PrivateSettings template

Universal-Omega:
GoogleMaps does not seem to work anymore as it is paid. Therefore this should be removed.


Reception123 (changes requested):
Also needs to be removed from https://github.com/miraheze/puppet/blob/master/modules/mediawiki/manifests/init.pp

Universal-Omega:
> Also needs to be removed from https://github.com/miraheze/puppet/blob/master/modules/mediawiki/manifests/init.pp

Oops. My bad. Fixed.

Reception123 (approved):

PR #1595, opened 2021-01-15
Remove `z.moatads.com` from CSP whitelist

dmehus:
Per @Southparkfan, I can't see a //need// to have `z.moatads.com` whitelisted. I wonder if a wiki requested this originally so they could serve ads? 🤔


dmehus:
Closed per discussion with @JohnFLewis; related to the AddThis extension. Feel free to reopen if we desire to remove the extension

PR #1600, opened 2021-01-22
[POC|DNM] data: apply changes from T6757

RhinosF1:


Southparkfan:
Besides the fact this reorganisation has not been ratified (yet), what is the reason for removing groups such as puppet-users?

modules/users/data/data.yaml
```diff
@@ -1,10 +1,10 @@
 groups:
-  ops:
+  infrastructure-engineers:
```
JohnFLewis, line 2: renaming groups is a manual process, personally I don't see why it should be renamed.

Also root on all server == site reliability engineers, they're not 'infrastructure engineers'

modules/users/data/data.yaml
```diff
@@ -16,26 +16,7 @@ groups:
                  'ALL = (ALL) NOPASSWD: /usr/sbin/service jobchron *',
                  'ALL = (ALL) NOPASSWD: /usr/bin/puppet *',
                  'ALL = (ALL) NOPASSWD: /root/ssl-certificate',
-                 'ALL = (ALL) NOPASSWD: /bin/journalctl *']
-  mediawiki-roots:
```
JohnFLewis, line 20: groups should really be kept even if unused, because the GIDs were in use and reusing GIDs is poor practise.

dmehus:
This seems a little like putting the cart before the horse, and the foal doesn't even have a name yet.

modules/users/data/data.yaml
RhinosF1: Your proposal splits us into 2 groups. My understanding is those with root everywhere would be infrastructure engineers.

modules/users/data/data.yaml
RhinosF1: Will handle later

RhinosF1:
> Besides the fact this reorganisation has not been ratified (yet), what is the reason for removing groups such as puppet-users?

It's a proof of concept and marked do not merge and based on a short chat in -staff.

modules/users/data/data.yaml
JohnFLewis: Split into two departments, Site Reliability Engineers is an appropriate job description of what the Infrastructure team are, it’s about keeping role descriptors relevant which is why MediaWiki Administrators are proposed to not be called ‘Administrators’ because that’s not a good descriptor. They also can’t really be called Site Reliability Engineers because they don’t have the wider toolkit to manage the core infrastructure behind the MediaWiki stack. They also can’t be called Software Engineers because most aren’t active MediaWiki code developers and don’t have a primary focus on code development over sysadmin work

RhinosF1: ops != sre but it's a point of discussion. If we wish to keep it stating 'ops' or change to 'sre' then that's not an issue, I've deliberately split the changes into multiple commits so individual components can be reverted. It's also a concept to show how it **might** look.

RhinosF1:
> This seems a little like putting the cart before the horse, and the foal doesn't even have a name yet.

That's why it's labelled concept in the title

modules/users/data/data.yaml
dmehus: For what it's worth, if this is about the commands for MirahezeBots to yelp for help, we should have a more generic command like `!ops` to refer to channel operators, inclusive of, generally but not exclusively, SRE, MediaWiki administrators, stewards, and Global ChanOps (to the extent they continue to be authorized by Group Contacts)

RhinosF1: This has nothing to do with that. The ops ping stuff for Miraheze bots will come soon but is blocked on me implementing sane memory-cached dynamic configuration dicts.

JohnFLewis:
A PoC for an unapproved idea that has some issues with it. The logistics can be dealt with when approved and further details are discussed and sorted out.

PR #1601, opened 2021-01-22
Add translate.yandex.net to CSP to maybe? make yandex translate work

Universal-Omega:


dmehus:
@Universal-Omega I have no objections to this, if this is the problem. Wasn't the issue with not having sufficient credits available for our API key or something, though?

R4356th:
> @Universal-Omega I have no objections to this, if this is the problem. Wasn't the issue with not having sufficient credits available for our API key or something, though?

It just has not been closed yet. Per RhinosF1, Yandex was working till last August when the key was valid.

PR #1602, opened 2021-01-26
Add The Internet Archive to CSP whitelist

dmehus:
Per https://phabricator.miraheze.org/T6771. Just needs @Reception123 either of @paladox, @NDKilla, @JohnFLewis, or @Southparkfan to 👍


Pix1234 (changes requested):
Needs to be *.archive.org

dmehus:
> Needs to be *.archive.org

I debated that, but I didn't see _any_ URLs for media files, archive files, etc. hosted on _Internet Archive_ on subdomains other than strictly [archive.org](https://archive.org/), so in this case, the root domain name should be sufficient. It can always be amended later to add a wildcard if we discover that they host files on domains other than the root domain name

Universal-Omega:
I don't personally think this is the best idea to allow this. Not every thing in archive.org is safe.

Universal-Omega:
Apologies. Caused merge conflict that now needs resolved.

Reception123:
@Universal-Omega Perhaps we should first wait for the discussion on Phabricator to come to an end, as it's not very clear if we'll be adding this or not. Currently, I'm not yet happy to give my approval.

PR #1635, opened 2021-02-06
rdns-check: change regex to not match nonexistent cp*

RhinosF1:
Current regex will match cp4,5,6,7,8,9 etc all which don't exist and being pointed at that caused at least one wiki to go down. We should alert to find others.


Universal-Omega:
Won't 0-9 still work?

RhinosF1:
> Won't 0-9 still work?

0-9 includes 4,5,6,7,8,9 which all don't exist in our DNS but see IRC as this may not work.

RhinosF1:
@Southparkfan says this will work when we remove old cps from the dns which will happen in less than a day so pointless.

PR #1664, opened 2021-02-26
Have MirahezeLogbot insert {{TOC right}} on every edit

dmehus:
Might be a better way to fix this, but if MirahezeLogbot inserts `{{TOC right}}` with every edit, it'll always be at the top of `Tech:Server admin log`

### DIFF
--- a/modules/irc/files/logbot/adminlog.py
+++ b/modules/irc/files/logbot/adminlog.py
@@ -33,7 +33,7 @@ def log(config, message, project, author):
     position = 0
     # Um, check the date
     now = datetime.datetime.utcnow()
-    logline = "* %02d:%02d %s: %s" % (now.hour, now.minute, author, message)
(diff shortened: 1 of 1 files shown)


dmehus:
Meh, I need to work on this a bit. Please **do not merge this yet**

RhinosF1 (changes requested):
We don't want on every edit. We just need it to fix the position of where it is to be above the new log entry.

dmehus:
> We don't want on every edit. We just need it to fix the position of where it is to be above the new log entry.

True, and this didn't take into account subsequent SAL log entries on the same calendar day. Should I close this and let you start over?

RhinosF1:
> > We don't want on every edit. We just need it to fix the position of where it is to be above the new log entry.
> 
> True, and this didn't take into account subsequent SAL log entries on the same calendar day. Should I close this and let you start over?

Yes

dmehus:
Will let @RhinosF1 do a new PR for this.

RhinosF1:
Created #1667

> Will let @RhinosF1 do a new PR for this.

PR #1674, opened 2021-03-01
postfix: Add query filter for ldap aliases

paladox:


modules/postfix/templates/virtual_alias_maps
```diff
@@ -4,3 +4,6 @@ bind_dn = cn=write-user,dc=miraheze,dc=org
 bind_pw = <%= @ldap_password %>
 search_base = ou=people,dc=miraheze,dc=org
 scope = sub
+
+query_filter = (&(objectClass=postfixUser)(mail=%u))
+result_attribute = maildrop
```
JohnFLewis, line 9: Why are you using mail?

paladox: This didn't work, I was trying to prevent errors :)

JohnFLewis: That still doesn’t answer the question?

paladox: I was following an example I found and thought that could help.

JohnFLewis: But we don’t use mail for postfix at all, even if this did ‘work’ it would be reverted.

PR #1680, opened 2021-03-07
Varnish: Add X-Robots-Tag http header (Should fix T6680)

Universal-Omega:


Southparkfan:
What about the localised namespace variants? E.g. 'Speciaal' (Dutch) or 'Spezial' (German) instead of 'Special' (English')?

Universal-Omega:
> What about the localised namespace variants? E.g. 'Speciaal' (Dutch) or 'Spezial' (German) instead of 'Special' (English')?

Well, I just used the same ones that are set to disallow via [robots.php](https://github.com/miraheze/puppet/blob/master/modules/mediawiki/files/robots.php#L12-L16), but any recommendation to do this for other variants, or would it have to be manually listed for every one?

PR #1705, opened 2021-03-13
Fix X-Robots-Tag

Universal-Omega:


modules/varnish/templates/default.vcl
```diff
@@ -402,7 +402,7 @@ sub vcl_deliver {
 		set resp.http.Age = 0;
 	}
 
-	if (req.url ~ "^(/w/api\.php|/w/index\.php\?title\=Special\:|/wiki/Special\:|/w/index\.php\?title\=Special%3A|/wiki/Special%3A)$") {
+	if (req.url ~ "^(/w/api\.php*|/w/index\.php\?title\=Special\:|/wiki/Special\:|/w/index\.php\?title\=Special%3A|/wiki/Special%3A).+$") {
```
paladox, line 405: shouldn't it be for api.php .*?

Universal-Omega: I think this will work for /w/api.php*** and /w/api.php whereas that (.*) would just be /w/api.php***, unless I'm mistaken?

PR #1718, opened 2021-03-18
puppetserver: +ssh key

RhinosF1:
This key should be generated from jbr3 and will be used to sync private keys from jbr3->puppet2 under a new system.

The jobrunner stuff will come shortly in a seperate pr.

It must only be available to puppet-users but most of the human work will soon be eliminated.

### DIFF
--- a/modules/puppetserver/manifests/init.pp
+++ b/modules/puppetserver/manifests/init.pp
@@ -256,5 +256,13 @@ class puppetserver(
         vars          => {
             tcp_port    => '8140',
         },
+    users::user { 'puppet-users':
+        ensure      => present,
+        uid         => 2004,
+        gid         => 2004,
+        ssh_keys    => [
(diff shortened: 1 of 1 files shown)


JohnFLewis:
I've spoken to Reception already and raised that I'd like this work to be put on hold as we might be dropping the concept of jobrunners soon which means we'd have to tackle this problem again but in a more complicated fashion shortly in the future.

RhinosF1:
> I've spoken to Reception already and raised that I'd like this work to be put on hold as we might be dropping the concept of jobrunners soon which means we'd have to tackle this problem again but in a more complicated fashion shortly in the future.

If the puppet code was designed well enough then whatever became the new server for renewing ssl certificates would take over.

JohnFLewis:
There wouldn’t be ‘one’ server doing this

PR #1725, opened 2021-04-01
Switch firejail to py3

RhinosF1:


RhinosF1:
Worked upstream so should I guess

PR #1746, opened 2021-04-24
Remove  wmgYandexTranslationKey

R4356th:
See miraheze/mw-config#3854 for context.


Universal-Omega (changes requested):
Also needs removed from https://github.com/miraheze/puppet/blob/3b0191964ddb0b311491911fa73c3d514d24dffb/modules/mediawiki/manifests/init.pp#L108

R4356th:
> 
> 
> Also needs removed from
> 
> https://github.com/miraheze/puppet/blob/3b0191964ddb0b311491911fa73c3d514d24dffb/modules/mediawiki/manifests/init.pp#L108

Done.

PR #1748, opened 2021-04-28
users: remove usage of the deprecated function is_array

MacFan4000:
the function was deprecated as of puppet 5


paladox:
Thanks!

PR #1750, opened 2021-05-07
Only allow system administrators to use logbot

Universal-Omega:


JohnFLewis:
There is no need to restrict the log bot. There has been no problem in 6 years, I don’t understand people’s obsessions with this?

Reception123:
My ultimate view based on the points expressed above is that unless there's a major troll issue, there's not a real need to do this and the current system has worked fine.

JohnFLewis:
This has been discussed numerous times, and I know for a fact you have been involved in at least one of these discussions.

Universal-Omega:
> This has been discussed numerous times, and I know for a fact you have been involved in at least one of these discussions.

Then I'm sorry but I don't recall. But truly I do value your advice and I'm truly sorry you thought just because you left I did this.

PR #1752, opened 2021-05-07
make requestwiki AI job memory 1750M

Reception123:


dmehus:
@Reception123 Oops, I meant to click "comment" instead of "approve." Should we test this first before merging, by using the `--parameter` that specifies a minimum memory requirement (like @JohnFLewis did before once or twice)?

Reception123:
> @Reception123 Oops, I meant to click "comment" instead of "approve." Should we test this first before merging, by using the `--parameter` that specifies a minimum memory requirement (like @JohnFLewis did before once or twice)?

The issue is I don't know how to do that because when a wiki status changes the job is triggered but goes directly to the jobrunner which makes it fail, so I can't try running it with the higher memory

JohnFLewis:
> > @Reception123 Oops, I meant to click "comment" instead of "approve." Should we test this first before merging, by using the `--parameter` that specifies a minimum memory requirement (like @JohnFLewis did before once or twice)?
> 
> The issue is I don't know how to do that because when a wiki status changes the job is triggered but goes directly to the jobrunner which makes it fail, so I can't try running it with the higher memory

Disabled jobrunner on jobrunner3

Reception123:
> > > @Reception123 Oops, I meant to click "comment" instead of "approve." Should we test this first before merging, by using the `--parameter` that specifies a minimum memory requirement (like @JohnFLewis did before once or twice)?
> > 
> > 
> > The issue is I don't know how to do that because when a wiki status changes the job is triggered but goes directly to the jobrunner which makes it fail, so I can't try running it with the higher memory
> 
> Disabled jobrunner on jobrunner3

Ah good point since the other mw servers also run jobs now.

Reception123:
It seems to work!

PR #1756, opened 2021-05-19
move bots to Libera Chat

Reception123:


Southparkfan:
What about modules/phabricator/templates/botconfig.json.erb?

Universal-Omega:
> What about modules/phabricator/templates/botconfig.json.erb?

Yes that needs done also. That one was also missed last time when changing bots from #miraheze to #miraheze-sre in #1639. I had to do that afterwords, so it should be done this time.

Universal-Omega (approved):
Looks good to me now

RhinosF1 (approved):

PR #1758, opened 2021-05-22
set depth to 1 instead of 5 for cloning core

Universal-Omega:
To hopefully fix puppet timeouts


paladox:
This will break when pulling commits fyi. Will cause history errors so when you deploy a new commit you will have to do it manually.

Reception123:
> This will break when pulling commits fyi. Will cause history errors so when you deploy a new commit you will have to do it manually.

Sorry about this, I accidentally merged the wrong PR (I would've consulted you first) but after merging I thought it would be okay after all. Will revert.

PR #1761, opened 2021-05-24
set test3 branch to REL1_36

Reception123:


RhinosF1:
Missing config switch too but #1735

RhinosF1 (changes requested):
-

PR #1789, opened 2021-06-17
Update default.vcl

Universal-Omega:
This should potentially fix [T7441](https://phabricator.miraheze.org/T7441).

If this does not fix it, we would have to see why the following is not called:
```vcl
sub recv_purge {
	if (req.method == "PURGE") {
		if (!client.ip ~ purge) {
			return (synth(405, "Denied."));
		} else {
			return (purge);
		}
	}
}
```
That is supposed to be called with [`call recv_purge;`](https://github.com/miraheze/puppet/blob/ea1e5bb/modules/varnish/templates/default.vcl#L306) from [`vcl_recv`](https://github.com/miraheze/puppet/blob/ea1e5bb/modules/varnish/templates/default.vcl#L305).

The only reason I can see why this would not be called is if [`client.ip`](https://github.com/miraheze/puppet/blob/ea1e5bb/modules/varnish/templates/default.vcl#L225) is not in the [`acl purge`](https://github.com/miraheze/puppet/blob/ea1e5bb/modules/varnish/templates/default.vcl#L110-L136). If that's the case I am not sure what `client.ip` is calling the purge request or what IP needs added to the purge access control level.

There is no guarantee that anything I've said here is accurate, and may be totally off-base.

### DIFF
--- a/modules/varnish/templates/default.vcl
+++ b/modules/varnish/templates/default.vcl
@@ -9,8 +9,8 @@
 # please contact Southparkfan <southparkfan [at] miraheze [dot] org>.
 
 # Marker to tell the VCL compiler that this VCL has been adapted to the
-# new 4.0 format.
-vcl 4.0;
+# new 4.1 format.
+vcl 4.1;
 
 import directors;
 import std;
--- a/modules/varnish/templates/default.vcl
+++ b/modules/varnish/templates/default.vcl
@@ -10,7 +10,7 @@
 
 # Marker to tell the VCL compiler that this VCL has been adapted to the
 # new 4.0 format.
-vcl 4.1;
+vcl 4.0;
 
 import directors;
 import std;
@@ -108,7 +108,9 @@ sub vcl_init {
 
 
 acl purge {
+	# localhost
 	"localhost";
+	"127.0.0.1";
 	# mw8
 	"51.195.236.221";
 	"2001:41d0:800:178a::7";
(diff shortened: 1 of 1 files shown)


Reception123:
@The-Voidwalker do you think you'd have time to review this?

The-Voidwalker:
It's a bit lower on the priority list. I'll get to it, but it will be a bit.

modules/varnish/templates/default.vcl
```diff
@@ -346,6 +348,17 @@ sub vcl_hash {
 	}
 }
 
+sub vcl_hit {
+	if (!obj.ttl > 0s) {
```
JohnFLewis, line 352: Why !obj.ttl?

modules/varnish/templates/default.vcl
```diff
@@ -346,6 +348,17 @@ sub vcl_hash {
...
+		return (pass);
+	}
+
+	# Force lookup if the request is a no-cache request from the client.
+	if (req.http.Cache-Control ~ "no-cache") {
+		return (miss);
+	}
+}
+
```
JohnFLewis, line 361: What if neither if is executed? This would abandon the request entirely rather than return a result.

JohnFLewis (changes requested):

modules/varnish/templates/default.vcl
Universal-Omega: It was originally based off quite a bit of research, but now I'm not sure the vcl_hit is even necessary here as it should be purging from recv_purge anyways and called from vcl_recv so I have actually no idea why it isn't properly working anymore.

Universal-Omega:
Unlikely this will fix the issue.

PR #1793, opened 2021-06-25
check_rdns: throw away anything after /

RhinosF1:
We have a path based redirect. Without messing around with checking schema's and stuff urlparse won't be useful. Just split it at / and chuck anything after away as you won't see a / in a url


The-Voidwalker:
This type of redirect is not valid in our setup, we'll need to come up with a different way to do this.

PR #1799, opened 2021-06-28
Whitelist *.archive.org

MusikAnimal:
This is used by the YouTube extension following https://github.com/miraheze/YouTube/pull/11


dmehus:
@MusikAnimal I suggested this be whitelisted awhile back and perhaps I didn't push strong enough, but your argument there I think is even stronger. :+1: from me.

MusikAnimal:
Is the concern about privacy? I would guess not considering we've whitelisted Google, YouTube, Facebook, Twitter, etc.

ugochimobi (approved):

Universal-Omega (approved):
Even though I did not want to have it done last time I do think the reason given this is reason enough to at least consider it under our new policy for CSP whitelisting.

R4356th:
Just noting that this was declined earlier due to very good reasons- https://phabricator.miraheze.org/T6771.

Reception123:
A new CSP policy is being worked on and this will be reviewed and approved/declined subject to the upcoming policy.

Universal-Omega:
> Just noting that this was declined earlier due to very good reasons- https://phabricator.miraheze.org/T6771.

Yes I agree, but I also agree that there are other good points regarding Google and Dropbox. If this is declined I hope we can review those as well and see if they are absolutely needed or if we can change that.

Universal-Omega:
Also, merge conflict needs resolved.

JohnFLewis:
@The-Voidwalker is handling the CSP policy implementation. Assigning this to him to ensure it has a responsible person attached.

Universal-Omega:
This will probably need redone since csp_whitelist no longer exists and has been replaced by csp.yaml.

Reception123:
Will have to close due to the unresovable conflict because of the change in the CSP system.

PR #1817, opened 2021-07-21
jobrunner3 cannot run intensive jobs

The-Voidwalker:
Much safer to run on jobrunner4, which has no critical processes


hieradata/hosts/jobrunner4.yaml
```diff
@@ -1,6 +1,7 @@
 users::groups:
   - mediawiki-admins
 jobrunner: true
+jobrunner::intensive: false
```
Universal-Omega, line 4: Shouldn't this be `true`?

PR #1821, opened 2021-07-26
add OAuth2 private key

Reception123:
fixes T7386


Reception123:
This doesn't make sense since the setting just points to a file path.

PR #1835, opened 2021-07-30
hieradata: set values for sync scripts

RhinosF1:
This means:
- by default, install the ssh key for www-data syncing and do not be a staging/canary server
- On test3, don't install the www-data key / be a staging server so it has the git clone but only skip syncing to other servers (/srv/mediawiki locally will still be setup and come from the local staging folder)
- On mw10, (this is random but it's a server and the 8/9 struggle more with load) install staging and canary (this means it will have the public & private key pair for access in a deploy keys folder and armed in ssh agent. Do not allow remote access using the ssh keys. By default, sync files from the deploy command to all mediawiki servers except test3

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -15,3 +15,5 @@ puppetserver_hostname: puppet3.miraheze.org
 role::salt::minions::salt_master: 'puppet3.miraheze.org'
 dns: false
 mediawiki::use_staging: false
+mediawiki::is_canary: false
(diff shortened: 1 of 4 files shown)


RhinosF1:
Need to fix this. Please don't merge yet though as it will probably break mw10 if I'd have done it right.

RhinosF1:
@Reception123

Reception123 (approved):

PR #1854, opened 2021-08-06
Remove $wgReCaptchaSiteKey from PrivateSettings

Universal-Omega:
This can be done just in LocalSettings.php, since the site key is not private, only the 'secret' key is 'secret'


JohnFLewis:
Can not be merged until https://github.com/miraheze/mw-config/pull/4038 is merged.

Though is there much point in doing this? As it would separate two variables across two separate deploy systems unnecessarily which is mainly what it was done this was initially.

Reception123:
per above, merging to facilitate testing for V3.

PR #1866, opened 2021-08-11
logsalmsg: keep $USER is $SUDO_USER blank

RhinosF1:
This means that logs from the automatic config sync (while they happen) should still keep 'www-data' rather than show '@test3' by itself

### DIFF
--- a/modules/base/files/logsalmsg
+++ b/modules/base/files/logsalmsg
@@ -7,4 +7,9 @@ else
   ACTOR=$USER
 fi
 
+if [ $ACTOR = '' ]
+then
+  ACTOR=$USER
+fi
+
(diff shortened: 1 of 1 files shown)


modules/base/files/logsalmsg
```diff
@@ -7,4 +7,9 @@ else
   ACTOR=$USER
 fi
 
+if [ $ACTOR = '' ]
```
Universal-Omega, line 10: Why do this instead of just `|| [ $USER = 'www-data' ]` above?

Universal-Omega (changes requested):

modules/base/files/logsalmsg
Universal-Omega: Unless my logic is off this wouldn't work anyways as this exact setting would be the default as matching the above statement anyways.

PR #1876, opened 2021-08-12
[POC] mwscript: add some logging for 'long' scripts

RhinosF1:
Applies to all foreachwiki scripts + importDump

### DIFF
--- a/modules/mediawiki/files/bin/mwscript.py
+++ b/modules/mediawiki/files/bin/mwscript.py
@@ -2,9 +2,12 @@
 # -*- coding: utf-8 -*-
 import sys
 import os
+long = 0
 if len(sys.argv) < 3:
     raise Exception("Not Enough Parameters")
 script = sys.argv[1]
+if script == 'importDump.php':
+    long = 1
 if len(script.split('/')) == 1:
     script = f'/srv/mediawiki/w/maintenance/{sys.argv[1]}'
 else:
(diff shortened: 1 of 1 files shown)


modules/mediawiki/files/bin/mwscript.py
```diff
@@ -2,9 +2,12 @@
 # -*- coding: utf-8 -*-
 import sys
 import os
+long = False
 if len(sys.argv) < 3:
     raise Exception("Not Enough Parameters")
 script = sys.argv[1]
+if script == 'importDump.php':
```
Universal-Omega, line 9: I understand this isn't the final version of this, but this is probably better:
```py
longScripts = ("deleteBatch.php", "importDump.php", "importImages.php", "etc...")
if script in longScripts:
	long = True
```

Universal-Omega (changes requested):

modules/mediawiki/files/bin/mwscript.py
```diff
@@ -30,6 +35,8 @@
 print(command)
 confirm = input("Type 'Y' to confirm: ")
 if confirm.upper() == 'Y':
+    if long:
+        os.system(f'{command} (START)'})
     if 'generate' in locals():
         os.system(generate)
     return_value = os.system(command)
```
Universal-Omega, line 42: Can we maybe add the time taken if `long = True` also?

modules/mediawiki/files/bin/mwscript.py
RhinosF1: Yes, needed to think about which scripts first would be in the list

modules/mediawiki/files/bin/mwscript.py
RhinosF1: Yes we can capture time to execute

modules/mediawiki/files/bin/mwscript.py
```diff
@@ -6,7 +6,7 @@
 if len(sys.argv) < 3:
     raise Exception("Not Enough Parameters")
 script = sys.argv[1]
-if script == 'importDump.php':
+if script in ['importDump.php', 'deleteBatch.php', 'importImages.php', 'rebuildall.php']:
```
Universal-Omega, line 9: Isn't a tuple slightly more efficient in this case? Correct me if I'm wrong though. Also can this be a variable to make changing it later on a bit easier?

Universal-Omega (changes requested):

modules/mediawiki/files/bin/mwscript.py
RhinosF1: honestly, no idea on tuple v list.

As to changing it later. I'll change a lot when I add python CI in like a week so not worried.

Universal-Omega: Can this be merged then if your not planning to change that yet?

RhinosF1: Yes

PR #1877, opened 2021-08-12
Create workflow.yml

RhinosF1:


RhinosF1:
Quite a few failures

Universal-Omega:
Is it possible to disable `line too long` with this? Those are quite difficult to fix I think, unless you want to work on it.

Universal-Omega (changes requested):
Can we add notify-irc to this? Also after initial issues are fixed can this be modified to only run when python files are changed?

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: ['3.7', '3.9']
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
```
Universal-Omega, line 14: What's the purpose? Is there actually submodules being checked out? If so then it should stay but if not this can likely be removed.

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,26 @@
+on:
```
Universal-Omega, line 1: Can we add a workflow name? Also, can this be changed to:
```yaml
on:
  create:
  push:
  pull_request:
    types: [opened, synchronize]
```
As those triggeres seem to work best for other CIs and best for `notify-irc` if we add that.

Universal-Omega (changes requested):

RhinosF1:
> Can we add notify-irc to this? Also after initial issues are fixed can this be modified to only run when python files are changed?

Yes and yes we should be able to match on *.py

.github/workflows/workflow.yml
RhinosF1: I copied from previous templates. No it is not required.

.github/workflows/workflow.yml
RhinosF1: We can

Universal-Omega:
> > Can we add notify-irc to this? Also after initial issues are fixed can this be modified to only run when python files are changed?
> 
> Yes and yes we should be able to match on *.py

Yes I know that is possible with workflows, and is what I was referring to. Thanks!

Universal-Omega:
* Can you rebase this so we can see what violations remain, after #1956?

PR #1881, opened 2021-08-14
redis: Save every 900s if one change is made rather than 60s

paladox:


JohnFLewis:
We don't really need to save things to disk anymore for jobrunner as it should be a perfectly stable service. If this change was merged though, it would only operate at best a 5 minute backlog, but ideally all jobs should be run within 5 minutes which would make this change useless in theory except for the purpose of re-running old jobs and discarding all new jobs.

PR #1883, opened 2021-08-16
db11/12/13: Change buffer pool instance to one including size

paladox:
This is being done as both are OOM'g and reading up on the docs the benefit of the buffer pool are undone when swapping which each server is doing.


paladox:
Doc: https://mariadb.com/kb/en/innodb-buffer-pool/#innodb_buffer_pool_size

The-Voidwalker:
+1 from me. If my read of the documentation is correct, a large buffer pool instance won't really do us much good besides eat up possible ram. It's probably not worth tuning (aside from avoiding OOMs) until we can run mariadb without swapping.

PR #1891, opened 2021-08-26
mediawiki: Install ploticus from debian rather than manually

paladox:


Reception123:
@paladox it seems that it's actually not available with buster so we'll have to go back to installing manually.

PR #1892, opened 2021-08-26
Update to composer v2

MacFan4000:
DO NOT MERGE until bullseye upgrades are done. Buster native packages are v1.


paladox:
I'm going to repurpose this so we can move forward... rather then relying on debian we can just install composer.phar to /usr/bin/composer.

MacFan4000:
Ok, I had done it as the Debian package because that’s how @RhinosF1 wanted to do it.

RhinosF1:
> Ok, I had done it as the Debian package because that’s how @RhinosF1 wanted to do it.

At some point we should clean this logic up yes.

PR #1909, opened 2021-08-29
db: Increase table_definition_cache and table_open_cache to 20000

paladox:


JohnFLewis:
@paladox what is happening with this?

paladox:
> @paladox what is happening with this?

I was just thinking about this, I think I'm going to try 20,000 first because I'm a little unsure if 50,000 would cause a OOM.

PR #1910, opened 2021-08-29
mwdeploy: run composer with --world

RhinosF1:


RhinosF1:
This will download composer to active directory so maybe hang on

PR #1911, opened 2021-08-29
mediawiki: Purge ocaml

paladox:


paladox:
See https://gerrit.wikimedia.org/r/c/operations/puppet/+/540154

PR #1919, opened 2021-08-31
Add PHP-FPM icinga check

Universal-Omega:
Implements check from https://github.com/regilero/check_phpfpm_status/blob/d2e29b0efecf27947408b4e2423fef9e0fb51908/check_phpfpm_status.pl


JohnFLewis:
Manually updating the list is not a good use of puppet’s power and creates manual work and technical debt.

Also; what additional monitoring does this provide? As it suggests it’s to do with work processes which is available via Grafana already

RhinosF1:
> Manually updating the list is not a good use of puppet’s power and creates manual work and technical debt.
> 
> Also; what additional monitoring does this provide? As it suggests it’s to do with work processes which is available via Grafana already

It's about alerting. The whole point of introducing the alert was to speed up our ability to check whether worker issues were the cause of other alerts. It makes far more sense to keep it with our current alerts than introduce a new system that requires me to look in more places.

JohnFLewis:
Grafana has more powerful alerting for data based monitoring. With the advent of introducing SLOs and SLIs, Grafana will play a bigger role in monitoring and alerting than it has in the past.

This shouldn’t require you to look in more places as you should already be looking at Grafana, it’s not a new system - it’s been in the infra for over 3 years

RhinosF1:
The new place isn't Grafana. It's emails for monitoring. So far we have IRC, Icinga's dashboard and Grafana's various dashboards. What I want is for alerts to be in a single place.

It's perfectly possible as I've seen it done to have data that's shown visually in grafana trigger an icinga alert if you'd rather than approach (which I suggested). The other option would be Prometheus alert manager as a replacement or grafana based IRC alerts to replace icinga ones.

JohnFLewis:
Given there exists this monitoring already in place, this won’t be merged as it duplicates existing information.

PR #1921, opened 2021-09-01
mediawiki: Update composer to version 2

paladox:


Universal-Omega:
Isn't this done with #1892?

paladox:
> Isn't this done with #1892?

Kind of, I think I'm going to do things differently. Buster uses version 1.8 though bullseye uses 2.0.

paladox:
Though I guess I can just repurpose his change so he gets some credit.

paladox:
#1892 1892

PR #1933, opened 2021-09-04
mediawiki and varnish: Tweek connections to handle more

paladox:
Also changes opcache settings for mediawiki.

### DIFF
--- a/modules/nginx/templates/nginx.conf.erb
+++ b/modules/nginx/templates/nginx.conf.erb
@@ -7,10 +7,10 @@ worker_processes <%= @nginx_worker_processes %>;
 worker_processes <%= scope.lookupvar('::virtual_processor_count') %>;
 <% end %>
 pid /run/nginx.pid;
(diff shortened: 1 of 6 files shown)


modules/nginx/templates/nginx.conf.erb
```diff
@@ -7,10 +7,10 @@ worker_processes <%= @nginx_worker_processes %>;
 worker_processes <%= scope.lookupvar('::virtual_processor_count') %>;
 <% end %>
 pid /run/nginx.pid;
-worker_rlimit_nofile 16192; # Twice the number of worker_connections
+worker_rlimit_nofile 20000; # Twice the number of worker_connections
```
Universal-Omega, line 10: This comment isn't accurate anymore.

Universal-Omega (changes requested):

paladox:
> @paladox What is happening with this?

I haven’t decided whether I want to do this, my original intention is to improve performance. I’m not exactly sure if it does that.

also whilst we’re here it would be great to have your opinion on it.

PR #1934, opened 2021-09-04
mwdeploy: add SSH keys back

Reception123:


RhinosF1:
Just needed an mkdir until staging patches get merged

PR #1945, opened 2021-09-04
re-write CSP to be more secure

JohnFLewis:


modules/varnish/manifests/init.pp
```diff
@@ -40,7 +40,7 @@
     
     $module_path = get_module_path($module_name)
 
-    $csp_whitelist = loadyaml("${module_path}/data/csp_whitelist.yaml")
+    $csp_whitelist = loadyaml("${module_path}/data/csp.yaml")
     $frame_whitelist = loadyaml("${module_path}/data/frame_whitelist.yaml")
```
Universal-Omega, line 44: frame_whitelist was deleted.

Universal-Omega (changes requested):

PR #1950, opened 2021-09-05
CSP: add embed.nicovideo.jp to `frame-src`

Universal-Omega:


Reception123:
(Temporary change until approved by CSP policy)

PR #1953, opened 2021-09-06
varnish: Use random over round_robin

paladox:


paladox:
Things I noticed is less cpu is used on mw8 and mw9.

paladox:
I’ve tested this a bit and seems to be fine… I can merge it later and monitor after.

PR #1960, opened 2021-09-07
CSP: add scratchblocks.github.io to `script-src` (T8002)

Universal-Omega:
Already in `img-src` so adding to `script-src` also.

I could not find a review task for this though. This can be a temporary measure until it's CSP is fully approved. It is possible I just missed the review task or it was missed during the mass creating of tasks.


Reception123:
Merging per reason given, will create a review task as well.

PR #1962, opened 2021-09-08
CSP: add openlayers.org to `script-src`

MusikAnimal:
This is needed by the Cargo extension to display OpenLayers maps


Universal-Omega (approved):
This is already in `img-src` so should be okay to do.

PR #1963, opened 2021-09-09
T8005: Use new 'ErrorPages' repository

Universal-Omega:


Universal-Omega:
As long as I did this PR correctly, it should be okay to merge, as 404.php and MissingWiki.php should both be operational in the new repository. It will make it possible to test changes from test3, to add localisation support, if this is merged.

PR #1970, opened 2021-09-14
test3: switch MediaWiki to REL1_37

Universal-Omega:


RhinosF1 (changes requested):
Let's wait for rc0 at least

Universal-Omega:
> Let's wait for rc0 at least

Yeah we can wait. This is just for whenever the time does come.

PR #1983, opened 2021-09-21
csp: add *.wikipedia.org to connect-src

R4356th:
A lot of user scripts and gadgets are loaded from Wikipedia. WMF Commons could be added here, too.

### DIFF
--- a/modules/varnish/data/csp.yaml
+++ b/modules/varnish/data/csp.yaml
@@ -143,3 +143,5 @@ connect-src:
   - '*.wikipedia.org'
   - 'www.mediawiki.org'
   - '*.wikimedia.org'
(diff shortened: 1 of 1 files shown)


R4356th:
> 
> 
> I don't know if `connect-src` is the proper one for this. I created tbe `connect-src` for wikidata.org only, I think for scripts script-src is appropriate, isn't it? Is there a specific CSP error in console mentioning `connect-src`?

Yes, there was an error mentioning connect-src, which restricts certain APIs including `XMLHttpRequest`, `WebSocket` etc.

Universal-Omega (approved):

Reception123:
@R4356th mind removing the extra line ^

PR #1992, opened 2021-09-22
Fix Flake8 violations

Universal-Omega:


modules/irc/files/logbot/adminlog.py
```diff
@@ -3,8 +3,8 @@
 import sys
 sys.path.insert(0, r'/etc/irclogbot/mwclient')
 
-import mwclient # noqa: E402
-import datetime # noqa: E402
+import mwclient  # noqa: E402
+import datetime  # noqa: E402
 
 sys.path.insert(0, r'/etc/irclogbot/mwclient')
```
The-Voidwalker, line 9: Isn't this a duplicate of line 4? Therefore we can probably remove lines 4 and 5, and then remove `# noqu: E402`?

Universal-Omega: We tried that last time, it caused it to break.

The-Voidwalker: Hmm, interesting, might want to look into why that is at some point, but for now we can leave it be if it works.

PR #2000, opened 2021-10-03
Enable tideways for CLI

Universal-Omega:
Currently it only effects fpm, this should effect CLI also if `use_tideways` is enabled, which it is on test3 only. This should allow using the `--profiler` command line option in maintenance scripts on test3.

After further looking at this, it appears to have not been used at all, so this actually uses `mediawiki::php::use_tideways`

### DIFF
--- a/modules/mediawiki/manifests/php.pp
+++ b/modules/mediawiki/manifests/php.pp
@@ -71,6 +71,7 @@ class mediawiki::php (
 
     php::extension { 'tideways-xhprof':
         ensure   => $profiling_ensure,
+        package_name => '',
         priority => 30,
         config   => {
             'extension'                       => 'tideways_xhprof.so',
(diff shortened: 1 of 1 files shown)


modules/mediawiki/manifests/php.pp
```diff
@@ -71,12 +71,10 @@
 
     php::extension { 'tideways-xhprof':
         ensure   => $profiling_ensure,
-        package_name => '',
```
paladox, line 74: Why is this being removed?

Universal-Omega: It didn't seem necessary. I can re-add it if it was.

paladox: Removing it makes it default to php-${title} and so it tries to install a package that isn’t in the apt repo.

Universal-Omega: Oh. OK, will re-add then. Thanks!

PR #2003, opened 2021-10-03
Remove parsoid.js from miraheze_project_sys.yaml

Universal-Omega:


Universal-Omega:
* #2036 would delete the entire file anyways.

PR #2029, opened 2021-10-14
Remove `mediawiki::cron`

Universal-Omega:
Per discussion with the MediaWiki team, LocalisationUpdate would likely be removed. Therefore, the cron is now unnecessary.


Universal-Omega:
> Why is LocalisationUpdate being removed? It seems like a rather sensible extension which is maintained and used by the WMF?

Because we don't even have it properly configured, or utilised. From what I can tell, it simply generates i18nupdate files, but is never actually used. It also seems to be disabled within WMF.

JohnFLewis:
We should properly utilise this otherwise we had i18n updates being massively backlogged. Or an alternative system should be proposed to bring forward core i18n changes as we assumed we had this process in place for the last 6 years.

JohnFLewis:
Core i18n updates?

PR #2107, opened 2021-11-05
Disallow indexing /w/ / /w/index.php?*

AgentIsai:
While reading documentation from MediaWiki.org regarding MediaWiki:Robots.txt, I stumbled upon a suggestion from the documentation on how to disallow pages starting in /w/index.php from being indexed if you have URL rewrites enabled. After checking the Google Search Console for miraheze.org, I saw that there are hundreds of thousands (millions?) of pages crawled by Google starting in "/w/index.php". Most if not all of these crawled pages starting in /w/index.php" clash with existing pages on /wiki/ and thus are not shown on Google meaning that Google has crawled thousands upon thousands of pages only to not show the pages served by index.php. All those thousands of extra hits surely cause undue burden to Miraheze's servers and end up not reaping any benefits if hidden away because Google correctly indexed pages in the /wiki/ directory of the wiki and then decided to not include the results from pages indexed via index.php.

The approach in my commit is one used by Wikimedia. I got it from the MediaWiki.org page "Manual:Robots.txt" and have verified against Wikipedia/Wikimedia Meta's robots.txt is the solution used by them to prevent unnecessary page hits from web crawlers. This will disallow all pages in /w/ from being indexed (including index.php) but allow for any necessary JS/CSS be loaded for things such as the Wayback Machine/other crawlers who need it to properly render the site.

For more info on this, see https://www.mediawiki.org/wiki/Manual:Robots.txt

### DIFF
--- a/modules/mediawiki/files/robots.php
+++ b/modules/mediawiki/files/robots.php
@@ -32,6 +32,11 @@ echo "# Block SemrushBot" . "\r\n";
 echo "User-Agent: SemrushBot" . "\r\n";
 echo "Disallow: /" . "\r\n\n";
 
+# Disallow indexing /w/
+echo "# Disallow indexing /w/index.php?" . "\r\n\n";
+echo "Allow: /w/load.php?" . "\r\n\n";
+echo "Disallow: /w/" . "\r\n\n";
+
 if ( $databasesArray['combi'] ) {
 	if ( preg_match( '/^(.+)\.miraheze\.org$/', $_SERVER['HTTP_HOST'], $matches ) ) {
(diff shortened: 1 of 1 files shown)


Universal-Omega (changes requested):
https://github.com/miraheze/puppet/blob/79340d40305c26ccad88f5286b98bbf54fa69ca1/modules/varnish/templates/default.vcl#L405 should be modified instead. Otherwise this will give indexed but disallowed by robots.txt errors I believe. That's why I removed disallow stuff from here previously and added that.

AgentIsai:
Per above

PR #2108, opened 2021-11-05
Noindex for /w/index.php?*

AgentIsai:
Per suggestion from @Universal-Omega (see https://github.com/miraheze/puppet/pull/2107). 

The goal is to prohibit Google/Bing from uselessly indexing /w/index.php and causing clashes with articles indexed via /wiki/ in search engines.

### DIFF
--- a/modules/varnish/templates/default.vcl
+++ b/modules/varnish/templates/default.vcl
@@ -447,7 +447,7 @@ sub vcl_deliver {
 		set resp.http.Age = 0;
 	}
 
-	if (req.url ~ "^(/w/api\.php*|/w/index\.php\?*|/wiki/Special\:|/wiki/Special%3A).+$") {
+	if (req.url ~ "^(/w/(api|index|rest)\.php*|/wiki/Special(\:|%3A)).+$") {
 		set resp.http.X-Robots-Tag = "noindex";
 	}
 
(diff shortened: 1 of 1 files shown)


modules/varnish/templates/default.vcl
```diff
@@ -447,7 +447,7 @@ sub vcl_deliver {
 		set resp.http.Age = 0;
 	}
 
-	if (req.url ~ "^(/w/api\.php*|/w/index\.php\?title\=Special\:|/wiki/Special\:|/w/index\.php\?title\=Special%3A|/wiki/Special%3A).+$") {
+	if (req.url ~ "^(/w/api\.php*|/w/index\.php\?*|/wiki/Special\:|/wiki/Special%3A).+$") {
```
Universal-Omega, line 450: Not `*` just `/w/index\.php\?` would be enough I think. Either that or maybe `/w/index\.php*`, which might be better?

RhinosF1: Do api/rest need indexing?

Universal-Omega: Api is already there.

RhinosF1: Cant api and index be merged into one (api|index|rest)\.php* statement?

Universal-Omega: Use `^(/w/(api|index|rest)\.php*|/wiki/Special(\:|%3A)).+$` I believe

RhinosF1 (approved):

Universal-Omega (approved):

PR #2114, opened 2021-11-07
deploy-mediawiki: add `--pull` argument

Universal-Omega:


modules/mediawiki/files/bin/deploy-mediawiki.py
```diff
@@ -33,12 +33,34 @@ def run(args, start):
         os.system(f'/usr/local/bin/logsalmsg {text}')
     else:
         print(text)
+    if args.world:
+        if args.pull:
+            pull = str(args.pull).split(',')
+            if 'world' not in pull:
+                pull.append('world')
+                args.pull = ','.join(pull)
+            pull = None
```
RhinosF1, line 42: Why reset pull here

modules/mediawiki/files/bin/deploy-mediawiki.py
```diff
@@ -33,12 +33,34 @@ def run(args, start):
...
+        else:
+            args.pull = 'world'
```
RhinosF1, line 44: should always be an array

modules/mediawiki/files/bin/deploy-mediawiki.py
```diff
@@ -33,12 +33,34 @@ def run(args, start):
         os.system(f'/usr/local/bin/logsalmsg {text}')
     else:
         print(text)
+    if args.world:
+        if args.pull:
+            pull = str(args.pull).split(',')
+            if 'world' not in pull:
+                pull.append('world')
+                args.pull = ','.join(pull)
```
RhinosF1, line 41: doesn't seem needed

modules/mediawiki/files/bin/deploy-mediawiki.py
```diff
@@ -33,12 +33,34 @@ def run(args, start):
...
+            pull = None
+        else:
+            args.pull = 'world'
+    if args.pull:
+        pull = str(args.pull).split(',')
+        if 'config' in pull:
+            os.chdir('/srv/mediawiki-staging/config')
+            exitcodes.append(os.system('sudo -u www-data git pull --recurse-submodules --quiet'))
```
RhinosF1, line 49: no submodules in this repo but if we're gonna use same command then stick it in a variable and don't repeat

modules/mediawiki/files/bin/deploy-mediawiki.py
```diff
@@ -33,12 +33,34 @@ def run(args, start):
...
+            if 'world' not in pull:
+                pull.append('world')
+                args.pull = ','.join(pull)
+            pull = None
+        else:
+            args.pull = 'world'
+    if args.pull:
+        pull = str(args.pull).split(',')
+        if 'config' in pull:
+            os.chdir('/srv/mediawiki-staging/config')
```
RhinosF1, line 48: why not git -C

RhinosF1 (changes requested):

modules/mediawiki/files/bin/deploy-mediawiki.py
Universal-Omega: args.pull is a string that's split, no?

modules/mediawiki/files/bin/deploy-mediawiki.py
Universal-Omega: How come? This is to set args.pull to append world if it doesn't exist already, if world is being ran, so that the following git pull commands are used for it.

modules/mediawiki/files/bin/deploy-mediawiki.py
Universal-Omega: I guess that isn't necessary yeah, it was mainly because this instance of pull is being conditionally set if world is used, whereas its set again under args.pull directly.

modules/mediawiki/files/bin/deploy-mediawiki.py
Universal-Omega: Good point.

modules/mediawiki/files/bin/deploy-mediawiki.py
Universal-Omega: Will do, thanks!

modules/mediawiki/files/bin/deploy-mediawiki.py
RhinosF1: but after this point you use the pull variable

modules/mediawiki/files/bin/deploy-mediawiki.py
RhinosF1: ah right!

modules/mediawiki/files/bin/deploy-mediawiki.py
Universal-Omega: Yeah, but only if world is used is pull already set, if it isn't pull isn't set, and that's why pull is redefined below, if it's redefined below without args.pull being set which is what this is for, then pull being used here has no effect has it'd be overwritten with args.pull.

modules/mediawiki/files/bin/deploy-mediawiki.py
```diff
@@ -33,12 +33,34 @@ def run(args, start):
         os.system(f'/usr/local/bin/logsalmsg {text}')
     else:
```
RhinosF1, line 34:
```suggestion
    else:
        print(text)
    if args.pull:
        pull = str(args.pull).split(',')
        if args.world not in pull:
            pull.append('world')
        if 'config' in pull:
```

modules/mediawiki/files/bin/deploy-mediawiki.py
RhinosF1: see code suggestion

modules/mediawiki/files/bin/deploy-mediawiki.py
Universal-Omega: This wouldn't work either, as then we are always pulling even if args.world isn't used. Though I guess `if args.world and 'world' not in pull:` might work?

RhinosF1 (approved):

PR #2121, opened 2021-11-11
mwscript: rewrite to use argparse

Universal-Omega:
* use argparse
* running scripts for specific extensions or skins is now done with `--extension` or `--skin` instead of input prompt
* maintenance script arguments are no longer required to be in quotations, all unknown arguments to mwscript are now used for the maintenance script arguments
* adds `--no-log` argument
* adds `--confirm` argument
* converts long scripts to a defined tuple, and adds support to add extension maintenance scripts to the tuple if it is necessary in the future
* adds `nukeNS.php` and `refreshLinks.php` to `longscripts`

### DIFF
--- a/modules/mediawiki/files/bin/mwscript.py
+++ b/modules/mediawiki/files/bin/mwscript.py
@@ -30,7 +30,7 @@ def run(args):
     else:
         command = f'sudo -u www-data php {script} --wiki={wiki}'
     if args.arguments:
-        command = f'{command} {args.arguments}'
+        command += ' ' + ' '.join(args.arguments)
     logcommand = f'/usr/local/bin/logsalmsg "{command}'
     print("Will execute:")
     if 'generate' in locals():
@@ -54,9 +54,11 @@ def run(args):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Process some integers.')
-    parser.add_argument('script', required=True)
-    parser.add_argument('wiki', required=True)
-    parser.add_argument('arguments', nargs='*')
+    parser.add_argument('script')
+    parser.add_argument('wiki')
+    parser.add_argument('arguments', nargs='*', default='')
     parser.add_argument('--extension', '--skin', dest='extension')
     parser.add_argument('--no-log', dest='nolog', action='store_true')
-    run(parser.parse_args())
+
+    args, args.arguments = parser.parse_known_args()
+    run(args)
--- a/modules/mediawiki/files/bin/mwscript.py
+++ b/modules/mediawiki/files/bin/mwscript.py
@@ -38,14 +38,15 @@ def run(args):
     print(command)
     confirm = input("Type 'Y' to confirm: ")
     if confirm.upper() == 'Y':
-        if long:
+        if long and not args.nolog:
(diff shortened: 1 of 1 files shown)


modules/mediawiki/files/bin/mwscript.py
```diff
@@ -1,46 +1,64 @@
...
+        print(generate)
+    print(command)
+    confirm = input("Type 'Y' to confirm: ")
+    if confirm.upper() == 'Y':
+        if long and not args.nolog:
+            os.system(f'{logcommand} (START)"')
+        if 'generate' in locals():
+            os.system(generate)
+        return_value = os.system(command)
+        logcommand = f'{logcommand} (END - exit={str(return_value)})"'
```
Universal-Omega, line 46: I think this can probably be changed to `logcommand += f' (END - exit={str(return_value)})"'`?

modules/mediawiki/files/bin/mwscript.py
```diff
@@ -1,46 +1,64 @@
...
-    logcommand = f'{logcommand} (END - exit={str(return_value)})"'
-    print(f'Logging via {logcommand}')
-    os.system(logcommand)
-    print('Done!')
-else:
-    print('Aborted!')
+        print(generate)
+    print(command)
+    confirm = input("Type 'Y' to confirm: ")
+    if confirm.upper() == 'Y':
```
Universal-Omega, line 40: We should probably add a `-y` (or similar) argument so that this can be bypassed. Otherwise the idea behind `--no-log` is pointless also, as we wouldn't be able to use mwscript from other scripts anyways.

Universal-Omega (commented):
Just some things to note for future or maybe can be added to this PR also.

modules/mediawiki/files/bin/mwscript.py
RhinosF1: That would be of 0 benefit

modules/mediawiki/files/bin/mwscript.py
RhinosF1: Sounds good

modules/mediawiki/files/bin/mwscript.py
```diff
@@ -1,46 +1,64 @@
...
+        long = True
+    if len(script.split('/')) == 1:
+        script = f'/srv/mediawiki/w/maintenance/{script}'
+    else:
+        scriptsplit = script.split('/')
+        script = f'/srv/mediawiki/w/{scriptsplit[0]}/{scriptsplit[1]}/maintenance/{scriptsplit[2]}'
+        if scriptsplit[2] in longscripts:
+            long = True
+
+    wiki = args.wiki
```
RhinosF1, line 22: Is this needed? Why not just use args.wiki

RhinosF1 (approved):
Can't see anything massive. One nit below.

modules/mediawiki/files/bin/mwscript.py
Universal-Omega: Because it's used multiple times this just seemed cleaner.

Reception123 (approved):

PR #2127, opened 2021-11-17
Handle CORS preflight requests + fix `Access-Control-Allow-Origin` for api.php

Universal-Omega:
Hopefully will fix T7311

Always setting `Access-Control-Allow-Origin = "*"` for api.php is not good practice, as it allows unauthenticated requests to the API from anywhere (and for requests with `Access-Control-Allow-Credentials: true`, and `credentials: 'include'` this would be a security risk if the browsers did not block this). This removes that while attempting to maintain the hack for https://phabricator.wikimedia.org/T217669.

Currently, for example any API requests like `
```js
fetch('https://login.miraheze.org/w/api.php', {
  mode: 'cors',
  credentials: 'include'
})
```
will fail because `Access-Control-Allow-Origin = "*"` is being set which the browser will block for security reasons when `credentials: 'include'`. I am hoping this will resolve all of these issues, while still maintaining the fix that it was originally intended for.

### DIFF
--- a/modules/varnish/templates/default.vcl
+++ b/modules/varnish/templates/default.vcl
@@ -255,19 +255,20 @@ sub vcl_synth {
 		set resp.http.Access-Control-Allow-Origin = "*";
 	}
 
-	if (req.http.Host == "static.miraheze.org") {
-		// Handle CORS preflight requests
-		if (resp.reason == "CORS Preflight") {
-			set resp.reason = "OK";
-			set resp.http.Connection = "keep-alive";
-			set resp.http.Content-Length = "0";
+	// Handle CORS preflight requests
+	if (
+		req.http.Host == "static.miraheze.org" &&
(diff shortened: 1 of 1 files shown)


modules/varnish/templates/default.vcl
```diff
@@ -247,6 +247,28 @@ sub vcl_synth {
...
 	}
+
+	if (resp.reason == "T217669") {
+		set resp.reason = "OK";
+		set resp.http.Connection = "keep-alive";
+		set resp.http.Content-Length = "0";
+		set resp.http.Access-Control-Allow-Origin = "*";
+	}
+
+	if (req.http.Host == "static.miraheze.org") {
```
JohnFLewis, line 258: Can this be dropped into an if and if?

Universal-Omega: Thanks; done

PR #2134, opened 2021-11-22
Remove `mwtask1_test` backend

Universal-Omega:
This is exactly a duplicate of `mwtask1` backend, since there is no health check on the `mwtask1` backend.


RhinosF1 (changes requested):
Diff doesn't look right

PR #2135, opened 2021-11-23
Fix undefined/api.php redirect

Universal-Omega:


modules/varnish/templates/default.vcl
```diff
@@ -321,7 +326,6 @@ sub mw_vcl_recv {
 
 	// HACK for phabricator.wikimedia.org/T217669
 	if (req.url ~ "/w(iki)?/undefined/api.php") {
```
paladox, line 328: Why not move this to the top?

Universal-Omega: Wouldn't doing returning the synthetic response before, prevent the backend from ever being set?

PR #2141, opened 2021-12-03
mediawiki: update to 1.37

RhinosF1:


hieradata/hosts/mw11.yaml
```diff
@@ -3,7 +3,7 @@ users::groups:
 jobrunner: true
 contactgroups: ['sre', 'mediawiki']
 role::mediawiki::use_strict_firewall: true
-mediawiki::branch: 'REL1_36'
+mediawiki::branch: 'REL1_37'
```
Universal-Omega, line 6: Can't this just be removed? It reads from common and to change just test3, it can be explicitly set there. Just makes only having to do one file each time for production upgrades.

DarkMatterMan4500:
MediaWiki 1.37, here we come.

PR #2177, opened 2021-12-29
mediawiki: add new servers

RhinosF1:
- set puppet to local DC
- jobrunner off until IPs given & jobchron
- 7.4 php on all
- mwtask101 configured for staging to SCSVG
- test configured for staging with no deploy

Please deploy to task last to avoid race errors


modules/mediawiki/files/bin/deploy-mediawiki.py
```diff
@@ -110,7 +110,7 @@ def run(args, start):
         serverlist = ['mw8', 'mw9', 'mw10', 'mw12', 'mw13', 'mwtask1']
         sync = True
     elif args.servers == 'scsvg':
-        serverlist = ['mw101', 'mw102', 'mw111', 'mw112', 'mw121', 'mw122', 'mw131', 'mw132']
+        serverlist = ['mw101', 'mw102', 'mw111', 'mw112', 'mw121', 'mw122']
```
Universal-Omega, line 113: What about mwtask101?

manifests/site.pp
```diff
@@ -72,7 +72,14 @@
...
 
-node 'mwtask1.miraheze.org' {
+# new hardware
+node /^mw(1[012][12])\.miraheze\.org$/ {
+    include base
+    include role::mediawiki
+    include prometheus::php_fpm
+}
+
+node 'mwtask(1|101).miraheze.org' {
```
Universal-Omega, line 82: This needs to be in regex format: `/^mwtask(1|101)\.miraheze\.org$/`. It could also be `mwtask(1(01)?)` if its better but probably more complicated.

manifests/site.pp
```diff
@@ -99,7 +106,7 @@
     include role::salt
 }
 
-node 'test3.miraheze.org' {
+node /^test(3|111).miraheze.org$/ {
```
Universal-Omega, line 109: This needs an escape on the `.`: `\.miraheze\.org`.

manifests/site.pp
```diff
@@ -72,7 +72,14 @@
     include prometheus::php_fpm
 }
 
-node 'mwtask1.miraheze.org' {
+# new hardware
+node /^mw(1[012][12])\.miraheze\.org$/ {
```
Universal-Omega, line 76: Can this not just be combined with the above with `/^mw(([89]|1[0123])|1[012][12])\.miraheze\.org$/`?

Universal-Omega (changes requested):

modules/mediawiki/files/bin/deploy-mediawiki.py
RhinosF1: It's the canary. Same as mw11 not being in the list for ovlon

Universal-Omega: Oh, right!

Universal-Omega (approved):

PR #2191, opened 2021-12-31
base::dns: Support using ipv6 address for the resolver

paladox:
This is needed for ipv6 only host otherwise the internet is broken.


JohnFLewis:
All hosts are IPv6, wouldn’t it be easier to make v6 the standard rather than use a hiera switch on every host defined, otherwise things will break?

PR #2211, opened 2022-01-01
check_reverse_dns: use `dns.resolver.Resolver.resolve()`

Universal-Omega:
Per:
```ps
/usr/lib/nagios/plugins/check_reverse_dns.py:101: DeprecationWarning: please use dns.resolver.Resolver.resolve() instead
```


paladox:
we can merge this when we are about to switch to the new cluster.

RhinosF1:
We need to use AAAA records too

RhinosF1:
I done ipv6 as #2212 as it should work on both clusters

PR #2218, opened 2022-01-01
Support mon111 in logsalmsg

paladox:


modules/base/files/logsalmsg
```diff
@@ -7,4 +7,8 @@ else
   ACTOR=$USER
 fi
 
-echo '!log' "[$ACTOR@$HOSTNAME] $*" | nc -q0 -u 51.195.236.249 5071
+if [ "$HOSTNAME" == "mon111" ]; then
```
JohnFLewis, line 10: All servers except mon111 would use mon2’s v4 address

paladox: Oh, right. Guess this will need a hira config.

PR #2250, opened 2022-01-03
irc: use irc.libera.chat rather than direct IPv6 address

Universal-Omega:
This should theoretically work now after the following commits:
* 99796ef
* cd1ea41

Which were done after these were switched to use direct IPv6 addresses.

Using irc.libera.chat does work for other bots like MirahezeLogbot so I'm thinking that with the above commits it should also work here.

### DIFF
--- a/modules/role/manifests/irc.pp
+++ b/modules/role/manifests/irc.pp
@@ -4,11 +4,7 @@ class role::irc {
 
     class { 'irc::ircrcbot':
         nickname     => 'MirahezeRC',
-        # irc.libera.chat (ipv6 address, have to hardcode it.
-        # This is because it's either picking up the ipv4 address
-        # for the hostname or it doesn't support getting the ipv6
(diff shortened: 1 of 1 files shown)


paladox:
This doesn't work hence why I had to do this.

Universal-Omega:
> This doesn't work hence why I had to do this.

I thought it was an attempt that didnt work which is why the mentioned commits ended up fixing. If not then I misunderstood and apologies for that.

dmehus:
> This doesn't work hence why I had to do this.

I'm guessing this might because our back-end infrastructure is using only IPv6 IPs possibly?

Universal-Omega:
> > This doesn't work hence why I had to do this.
> 
> I'm guessing this might because our back-end infrastructure is using only IPv6 IPs possibly?

Yes we know that. Alot of things needed switching to support IPv6.

PR #2252, opened 2022-01-04
mwupgradetool: switch to SCSVG servers

Universal-Omega:


RhinosF1:
This can in theory be used whenever. We only use this once every 6 months.

PR #2261, opened 2022-01-05
use sslname if present for in checking reverse_dns

Universal-Omega:


RhinosF1:
Why

Universal-Omega:
> Why

Because for redirects that redirect the root domain to www for example, alert, when unnecessary, since they work, but alot of the time cnames can't be set for root domains in the same way, which then alert for the root domain redirect. If this isn't necessary or wrong approach can be closed though.

PR #2279, opened 2022-01-07
Revert all OOM mitigations

Universal-Omega:
Timeline:
* Servers were OOMing
* These changes were made
* OOMs reduced
* user-facing errors become 10x worse (502s/503s)

This was not the right fix (so it would seem), an investigation to what was causing the memory leak to begin with needs to be conducted. Not fixes that unintentionally create additional user-facing issues as a side effect. If this PR is merged and user-facing errors do not improve it can be reverted again.

Doing this after discussion about the timeline with @AgentIsai, waiting for feedback from @Reception123 in the conversation still.

### DIFF
--- a/hieradata/hosts/test101.yaml
+++ b/hieradata/hosts/test101.yaml
@@ -10,8 +10,7 @@ php::php_version: '7.4'
 mediawiki::php::use_tideways: true
 nginx::use_graylog: true
 puppetserver_hostname: 'puppet111.miraheze.org'
-mediawiki::php::fpm::childs: 12
-mediawiki::php::fpm::fpm_max_memory: 256
+mediawiki::php::fpm::childs: 32
 mediawiki::use_staging: true
 mediawiki::remote_sync: false
 
--- a/modules/php/manifests/fpm/pool.pp
+++ b/modules/php/manifests/fpm/pool.pp
@@ -54,7 +54,7 @@ define php::fpm::pool(
         'listen.backlog' => 256,
         'pm'     => 'static',
         'pm.max_children' => $facts['virtual_processor_count'],
-        'pm.max_requests' => 1000,
+        'pm.max_requests' => 5000,
         'pm.status_path' => '/php_status',
         'access.format'  => '%{%Y-%m-%dT%H:%M:%S}t [%p] %{microseconds}d %{HTTP_HOST}e/%r %m/%s %{mega}M',
         'process.dumpable' => yes,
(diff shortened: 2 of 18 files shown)


Reception123:
I'd like someone from Infra to comment as well but as far as I'm concerned I would agree that these mitigations aren't a replacement for finding the cause of the memory leak.

modules/php/manifests/fpm/pool.pp
```diff
@@ -54,7 +54,7 @@
         'listen.backlog' => 256,
         'pm'     => 'static',
         'pm.max_children' => $facts['virtual_processor_count'],
-        'pm.max_requests' => 1000,
+        'pm.max_requests' => 10000,
```
paladox, line 57: Should leave this at a 1000?

hieradata/hosts/test101.yaml
```diff
@@ -10,8 +10,7 @@ php::php_version: '7.4'
 mediawiki::php::use_tideways: true
 nginx::use_graylog: true
 puppetserver_hostname: 'puppet111.miraheze.org'
-mediawiki::php::fpm::childs: 12
-mediawiki::php::fpm::fpm_max_memory: 256
+mediawiki::php::fpm::childs: 32
```
paladox, line 13: Shouldn't have 32 childs. It doesn't have the same ram as the mw*. Same fro test3

paladox (changes requested):

RhinosF1:
I'd already raised concerns about reverting this as it will block the new cluster.

The issues @Universal-Omega is concerned about with specific pages not loading consistently are not down to worker count.

It's likely a memory issue which is why I opened https://github.com/miraheze/mw-config/pull/4349 as the RfC and archives not loading on meta all have Module:Reply_to involved.

We need to look at logs for the specific pages with consistent issues and look for either a timeout on a subcall (parsoid for DT, I did do https://github.com/miraheze/mw-config/commit/d1c32a8dc8de320b4791a2c7b692355a86ce2596 last week to bring it way under our limits if parsoid became slow for a 502) or a memory issue.

Also regarding some of the specific pages also not loading on SCSVG, subcalls to parsoid will still use ovlon and so will anything like instant commons.

modules/php/manifests/fpm/pool.pp
Universal-Omega: The PR just fully reverts them to what they were prior.

hieradata/hosts/test101.yaml
Universal-Omega: The PR just fully reverts them to what they were prior.

PR #2298, opened 2022-01-11
varnish: allow proxying beta

RhinosF1:


modules/varnish/templates/default.vcl
```diff
@@ -529,6 +529,11 @@ sub vcl_recv {
 		}
 	}
 
+        if (req.http.Host ~ "^*.betaheze.org") {
```
JohnFLewis, line 532: This is invalid regex

JohnFLewis (changes requested):

modules/varnish/templates/default.vcl
Universal-Omega: This should be `^(.*\.)?betaheze\.org` in order to also do betaheze.org itself, or `^.*\.betaheze\.org` for just subdomains. I'm not certain if the periods need escaped or not though.

JohnFLewis: > I'm not certain if the periods need escaped or not though.

They don't *strictly* need to be, but should be.

Universal-Omega:
Conflicts

modules/varnish/templates/default.vcl
```diff
@@ -263,6 +263,11 @@ sub vcl_recv {
 		}
 	}
 
+  if (req.http.Host ~ "^(.*\.)?betaheze\.org") {
```
Universal-Omega, line 266: Tabbing

Universal-Omega (changes requested):

Reception123 (approved):

PR #2299, opened 2022-01-11
varnish: don't pass cookies to backend for load.php

RhinosF1:


Universal-Omega:
There is no cookies set for load.php is there?

Universal-Omega:
> `req.url !~ "^/w/load\.php"`
Is in the if statement, therefore the `else` is called for that, calling `mw_stash_cookie` which removes the cookie values it seems.

RhinosF1:
> There is no cookies set for load.php is there?

Any cookie is set on the domain

> > `req.url !~ "^/w/load\.php"`
> 
> Is in the if statement, therefore the `else` is called for that, calling `mw_stash_cookie` which removes the cookie values it seems.

It's an or statement that includes session cookies causing cache to be bypassed. This allows it to be cached safely.

Universal-Omega:
OK well what exact is the point of doing this, any real benefit then?

RhinosF1:
> OK well what exact is the point of doing this, any real benefit then? I may not be fully understanding though.

To serve cached resources from load.php to logged in users

Universal-Omega:
OK sounds good then. I understand what you mean now, I was just not understanding before. Thanks!

RhinosF1:
Should we do this?

modules/varnish/templates/default.vcl
```diff
@@ -275,6 +275,10 @@ sub mw_stash_cookie {
 }
 
 sub mw_evaluate_cookie {
+        if (req.url ~ "^/w/load\.php") {
+           unset req.http.Cookie; # load.php URLs should be specfic and sessions disabled, we can ignore cookies
```
JohnFLewis, line 279: can this not use mw_stash_cookie like below? Also is this not triggered by the below?

Universal-Omega: That's what I thought also but was told above that it wasn't.

JohnFLewis:
I am looking at overhauling major parts of caching anyway, so closing for now.

PR #2341, opened 2022-01-16
MediaWiki: support control groups

Universal-Omega:
NOTE: this will require servers to be rebooted.


modules/mediawiki/templates/initscripts/mw-cgroup.systemd.erb
```diff
@@ -0,0 +1,20 @@
+# Systemd job for MediaWiki's cgroup config
+# This file is managed by Puppet.
+[Unit]
+Description="Mw-cgroup"
+Before=hhvm.service
```
RhinosF1, line 5: Not hhvm

PR #2367, opened 2022-01-19
Remove support for ufw

Universal-Omega:
We now use ferm everywhere as far as I know. If I'm wrong, or if this is not something that is wanted to do at this time, feel free to close this PR.


Universal-Omega:
Actually took a second look and realised cloud only has support not actually using it yet. Apologies for doing this prematurely.

PR #2369, opened 2022-01-19
bastion: move firewall rule to role class

supertassu:
Moves the bastion public ssh access firewall rule to the bastion role.
base::firewall should not set up firewall rules for individual services,
that responsibility should be left to the puppet modules configuring
those individual services.

Also clarify naming in the bastion role to make the names more clear.


JohnFLewis:
Thank you for this, I should have done this initially!

PR #2409, opened 2022-01-31
postgresql::postgis: support Bullseye

Universal-Omega:


JohnFLewis:
@paladox can you look as the one who mostly maintains the postgres install here?

PR #2413, opened 2022-02-03
base::puppet: remove temporary Buster; use Bullseye

Universal-Omega:


RhinosF1:
Is this now supported?

Universal-Omega:
> Is this now supported?

Should be.

RhinosF1 (approved):
http://apt.puppetlabs.com/pool/bullseye/puppet7/p/

Universal-Omega:
Per IRC:
> 10:23 AM <+ CosmicAlpha> RhinosF1: http://apt.puppetlabs.com/pool/bullseye/puppet7/p/ — https://github.com/miraheze/puppet/pull/2188 says it didn't support for server or DB is why, but that seems to indicate it now does I think.

JohnFLewis:
It’s annoying support was added literally days after migration, but at least support exists now.

PR #2423, opened 2022-02-16
Allow static access from same domain

Universal-Omega:
* Allows static files to be accessed directly from the wiki subdomain or custom domain using /static.
  * Should exclude private access, but this should be verified.
* Allows access to favicon/apple touch icon directly from /favicon.ico or /apple-touch-icon.png.
* Rewrites sitemap.php to use MediaWiki's HttpRequestFactory in order to retrieve the sitemap directly from static, rather than redirecting to it. This is designed to fix support for the sitemap index usage within Google, etc...
  * Also provides better code standards rather than again decoding the database file, it uses WebStart, and then `$wgDBname` directly.
* Simplifies robots.php to use `$wgDBname` directly, rather than decoding the database file again. This should be accessible since we use WebStart.
* The use of WebStart for robots.php and sitemap.php should also add support for the beta realm, which uses a different suffix.

Meant to fix issues like https://phabricator.miraheze.org/T8744

I have no idea if this will work, or if my regex for static access is correct at all. This should be verified.

### DIFF
--- a/modules/mediawiki/files/favicon.php
+++ b/modules/mediawiki/files/favicon.php
@@ -1,0 +1,45 @@
+<?php
+
+// Based on the version created by Wikimedia
+
+define( 'MW_NO_SESSION', 1 );
+require_once '/srv/mediawiki/w/includes/WebStart.php';
+
+use MediaWiki\MediaWikiServices;
+
+function streamFavicon() {
+	global $wgFavicon;
+	wfResetOutputBuffers();
+
+	$favicon = $wgFavicon;
+	if ( $favicon === '/favicon.ico' ) {
+		$favicon = '/default.ico';
+	}
+
+	$req = RequestContext::getMain()->getRequest();
+	if ( $req->getHeader( 'X-Favicon-Loop' ) !== false ) {
+		header( 'HTTP/1.1 500 Internal Server Error' );
+		return;
+	}
+
+	$url = wfExpandUrl( $favicon, PROTO_CANONICAL );
+	$client = MediaWikiServices::getInstance()
+		->getHttpRequestFactory()
+		->create( $url );
+	$client->setHeader( 'X-Favicon-Loop', '1' );
+
+	$status = $client->execute();
+	if ( !$status->isOK() ) {
+		header( 'HTTP/1.1 500 Internal Server Error' );
+		return;
+	}
+
+	$content = $client->getContent();
+	header( 'Content-Length: ' . strlen( $content ) );
+	header( 'Content-Type: ' . $client->getResponseHeader( 'Content-Type' ) );
+	header( 'Cache-Control: public' );
+	header( 'Expires: ' . gmdate( 'r', time() + 86400 ) );
+	echo $content;
+}
+
+streamFavicon();
--- a/modules/mediawiki/files/favicon.php
+++ b/modules/mediawiki/files/favicon.php
@@ -13,7 +13,7 @@ function streamFavicon() {
 
 	$favicon = $wgFavicon;
 	if ( $favicon === '/favicon.ico' ) {
-		$favicon = '/default.ico';
+		$favicon = '/favicons/default.ico';
 	}
 
 	$req = RequestContext::getMain()->getRequest();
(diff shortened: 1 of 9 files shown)


Reception123:
@JohnFLewis Mind reviewing this?

JohnFLewis:
What would be the use of /static/? It seems like this double our cache need for static images for no tangible benefit?

The rest looks good though.

Universal-Omega:
> What would be the use of /static/? It seems like this double our cache need for static images for no tangible benefit?
> 
> The rest looks good though.

Additional usage for stuff like favicons, as currently favicon.ico with this won't work for setting $wgFavicon to it, unless we change $wgFavicon in ManageWiki to $wmgFavicon, and hard set $wgFavicon to /favicon.ico, while changing this PR to use $wmgFavicon for favicon.ico. The same would need done for the apple touch icon.

There's also other benefits to be able to load images from the same domain as well.

Universal-Omega:
@JohnFLewis any further thoughts on the above? Should I revert that portion of this PR?

Reception123:
T8744#179035 is quite curious - how could it have fixed itself?

dmehus:
What would be the reason to allow wikis to use their custom domains instead of `static.miraheze.org`? I personally think this makes it more cumbersome to easily identify wiki file servers in the format of `static.miraheze.org/dbname`. Will `static.miraheze.org/dbname` be retained as an _alias_ such that we can use either `static.customdomain.tld` _or_ `static.miraheze.org/dbname`? If _so_, then I'd have no real opposition to this, other than noting John's comment above

Reception123:
> What would be the reason to allow wikis to use their custom domains instead of `static.miraheze.org`? I personally think this makes it more cumbersome to easily identify wiki file servers in the format of `static.miraheze.org/dbname`. Will `static.miraheze.org/dbname` be retained as an _alias_ such that we can use either `static.customdomain.tld` _or_ `static.miraheze.org/dbname`? If _so_, then I'd have no real opposition to this, other than noting John's comment above

See the task I linked above for the original rationale, however I'm not sure if we need this now given that it was fixed by itself, but we should figure out why.

dmehus:
> T8744#179035

Ah, yes, indeed, that's interesting the original rationale reason (i.e., wiki favicons not loading correctly) has resolved itself. Definitely worth investigating.

Linking the Phabricator comment [here](https://phabricator.miraheze.org/T8744#179035)

Universal-Omega:
> What would be the reason to allow wikis to use their custom domains instead of `static.miraheze.org`? I personally think this makes it more cumbersome to easily identify wiki file servers in the format of `static.miraheze.org/dbname`. Will `static.miraheze.org/dbname` be retained as an _alias_ such that we can use either `static.customdomain.tld` _or_ `static.miraheze.org/dbname`? If _so_, then I'd have no real opposition to this, other than noting John's comment above

I never once mentioned static.customdomain.tld. this has nothing to do with that. It allows file access from wiki.domain.tld/static. While yes still retaining static.miraheze.org.

Universal-Omega:
> T8744#179035 is quite curious - how could it have fixed itself?

There may be some circumstances it would work. But unreliable and I can confirm the issue is present on my own wikis as well.

JohnFLewis:
> @JohnFLewis any further thoughts on the above? Should I revert that portion of this PR?

No further thoughts, rationale seems valid and my question was perfectly answered.

We can monitor cache hit levels once merged to see if there is any noticeable impact but I imagine there won’t be.

revi:
[Naver](https://en.wikipedia.org/wiki/Naver)'s Search Engine Console requires all links in the sitemap.xml to be within the same domain as the user's domain (in other words, links in https://revi.wiki/sitemap.xml has to be in https://revi.wiki/*) so loading it from revi.wiki/static/(whatever xml file) would solve my discoverability issue. (Basically a long `+1`)

Universal-Omega:
> [Naver](https://en.wikipedia.org/wiki/Naver)'s Search Engine Console requires all links in the sitemap.xml to be within the same domain as the user's domain (in other words, links in https://revi.wiki/sitemap.xml has to be in https://revi.wiki/*) so loading it from revi.wiki/static/(whatever xml file) would solve my discoverability issue. (Basically a long `+1`)

This change not only allows everything to be accessed via /static, but sitemaps (the sitemap index) in particular should actually load from the root sitemap.xml now as well. So shouldn't even need to use /static for sitemaps with these changes.

PR #2459, opened 2022-04-01
fix path to logsalmsg

MacFan4000:


dmehus (commented):
Thanks, @MacFan4000, for fixing this. I'll take your word for it on the updated path being correct. :)

PR #2460, opened 2022-04-03
mwtask: tweak php settings due to low traffic

RhinosF1:


paladox:
I don't see the benefit in doing this?

Universal-Omega:
What is the benefit of doing this?

PR #2465, opened 2022-04-05
jobs: move around

RhinosF1:
- stop scheduling RecordLintJob on high intensity, it's doing ok as a normal server I think.
- move all refreshLinks jobs to high intensity

### DIFF
--- a/modules/mediawiki/templates/jobrunner-hi.json.erb
+++ b/modules/mediawiki/templates/jobrunner-hi.json.erb
@@ -11,7 +11,6 @@
                 "RequestWikiAIJob",
                 "webVideoTranscode",
                 "webVideoTranscodePrioritized",
-                "RecordLintJob",
                 "refreshLinks"
             ]
         }
--- a/modules/mediawiki/templates/jobrunner.json.erb
+++ b/modules/mediawiki/templates/jobrunner.json.erb
@@ -14,7 +14,8 @@
                 "DataDumpGenerateJob",
                 "RequestWikiAIJob",
                 "webVideoTranscode",
-                "webVideoTranscodePrioritized"
+                "webVideoTranscodePrioritized",
+                "refreshLinks"
             ]
         }
     },


MacFan4000 (approved):

JohnFLewis:
refreshLink isn’t really a highly intensive job - all this will do is make it so it’s impossible to keep up with refreshLink jobs as some HI jobs can take hours to finish, which during that time you could easily get a few thousand (if not tens of thousands) jobs. This seems very unwise to me.

RhinosF1:
@JohnFLewis: A significant proportion were never finishing due to needing too much memory so they are running now on both sets.

I don't really fancy giving production more memory and I'm not sure whether having it in 2 sets is appropriate.

I'll admit to not currently being convinced about any solution.

RhinosF1:
That being in the current situation we were getting over a thousand build up any needing intervention anyway without them being allowed to run with higher memory so if task got locked up and busy for a while we could therefore end up in a shit situation again.

RhinosF1:
@JohnFLewis: any thoughts on what might work?

JohnFLewis:
Is there any reason why refreshLinks now requires substantially more memory than it has for the preceding year or two before? I think that is a problem that needs investigation more so than a bandage fix that’s ineffective to the root problem.

At a push, you can increase memory limits on the normal job runners for refreshLinks, otherwise you’d need more intensive job runners but there’s no appropriate candidates to locate them.

RhinosF1:
> Is there any reason why refreshLinks now requires substantially more memory than it has for the preceding year or two before? I think that is a problem that needs investigation more so than a bandage fix that’s ineffective to the root problem.

It's a bit of a concern. The issues started around March 30th. We're probably looking at something done on 29th or 30th. I've not figured what yet.

> At a push, you can increase memory limits on the normal job runners for refreshLinks, otherwise you’d need more intensive job runners but there’s no appropriate candidates to locate them.
That's definitely a push. You can bump the runners on in the config too rather than more VMs but given the load on task in the last day, not sure that would work either. Kinda expected there is no good solution as the answer.

JohnFLewis:
> improve page rendering performance by giving mediawiki more cpu + memory.

In the past I did offer more of both of these resources but they were refused and now we don’t have sufficient of either.

The problem can’t just be ‘it needs more resources’, as that’s how Orain began to collapse in financial ruin by throwing money at a problem rather than fixing it.

RhinosF1:
> > improve page rendering performance by giving mediawiki more cpu + memory.

It's slapping us in the face now with issues to bite us back.

> In the past I did offer more of both of these resources but they were refused and now we don’t have sufficient of either.

That was a call not looking far enough into the future

> The problem can’t just be ‘it needs more resources’, as that’s how Orain began to collapse in financial ruin by throwing money at a problem rather than fixing it.

mw* have high load nearly always anyway. This is icing on the cake to be honest or the tip of an iceberg. It's not the only evidence for more resources. We simply don't have the capacity to either render the pages or cope with the backlog by then failing and waiting to be given up on.

JohnFLewis:
It’s unfortunate but the state of affairs we’re in is that resources are now hitting the cap because we were expansive with them when we started. Looking into poor performing software is the only realistic way to resolve problems at the minute. Only way we will get more resources to throw st the problem is if we receive an average donation value of £7500 a year. Unfortunately we barely get half of this currently.

So the onus must change or we just need to accept the fact we can’t cope and stop trying to mitigate it by further straining resources.

PR #2476, opened 2022-04-07
logbot: Use OAUTH for login

paladox:
https://mwclient.readthedocs.io/en/latest/user/connecting.html#oauth


RhinosF1 (commented):
looks good but my really awkward style brain says brackets should open & close on the same column

PR #2488, opened 2022-04-16
switch from ordered_yaml to to_yaml

paladox:


Universal-Omega (changes requested):
`modules/vmlib/lib/puppet/functions/ordered_yaml.rb` should probably be deleted, and `ordered_yaml` removed from the vmlib README, since after this it is no longer used at all, we don't need to keep it.

PR #2517, opened 2022-04-22
varnish: add mw*_test without probe or pool

paladox:


JohnFLewis:
What problem does this solve? I intentionally removed them as part of redesigning how probes work

paladox:
> What problem does this solve? I intentionally removed them as part of redesigning how probes work

I was trying to fix blackbox. Discovered needing to open a firewall rule for prometheus to connects to mw* https port.

PR #2519, opened 2022-04-29
remove addthis.com from the CSP (per replacement, see T7876)

Reception123:
DO NOT MERGE until T7876 is completed.


Reception123:
whoops, didn't realise that branch actually existed. Since this is stalled, might as well close for now...

PR #2532, opened 2022-05-09
switch SSL certificate generation to puppet111/automate private keys [DO NOT MERGE!]

Reception123:
As explained in T7582, since mw-admins are not actively generating certs and to make things easier for myself and MacFan4000 (as well as the later automation) it makes much more sense to move the entire generation process to puppet111 and automate the SSL key part rather than keep it on mwtask111.

### DIFF
--- a/hieradata/hosts/mwtask111.yaml
+++ b/hieradata/hosts/mwtask111.yaml
@@ -1,7 +1,6 @@
 users::groups:
   - mediawiki-admins
   - mediawiki-roots
-letsencrypt: true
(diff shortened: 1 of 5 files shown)


Universal-Omega (changes requested):
https://github.com/miraheze/puppet/blob/ece79368c7dcb0ed54d6ac0614e14b1f7b63a4b9/modules/monitoring/files/scripts/ssl-renew.sh#L24

Don't forget to change the IP there I think.

PR #2537, opened 2022-05-13
Rewrite PHP configuration

Universal-Omega:
Base values appears to always be present over the individual modules (overriding them); configure PHP directly in their relevant modules with their respective configurations, and allow further easier configuration for MediaWiki PHP. This also allows us to only use php extensions necessary for the relevant service.

Also this uses dynamic all modules except for MediaWiki.

Inspired by Wikimedia's config, but adapted to fit our needs for our defaults, and our services.


modules/mediawiki/manifests/php.pp
```diff
@@ -1,47 +1,185 @@
...
+                'emergency_restart_threshold' => $facts['virtual_processor_count'],
+                'process.priority'            => -19,
+            }
+        }
+
+        # This will add an fpm pool listening on port 8000
+        # We want a minimum of 4 workers
+        $num_workers = max(floor($facts['virtual_processor_count'] * $fpm_workers_multiplier), 4)
+        php::fpm::pool { 'www':
+            port   => 8000,
```
paladox, line 171: Port 8080? We use unix sockets as this would be slower.

Universal-Omega: Fixed. Thanks!

PR #2541, opened 2022-05-14
Fallback to default icons if bad status is given

Universal-Omega:


modules/mediawiki/files/favicon.php
```diff
@@ -30,8 +30,17 @@ function streamFavicon() {
 
 	$status = $client->execute();
 	if ( !$status->isOK() ) {
-		header( 'HTTP/1.1 500 Internal Server Error' );
-		return;
+		$favicon = '/favicons/default.ico';
```
RhinosF1, line 33: Any need to assign for a single use?

modules/mediawiki/files/favicon.php
```diff
@@ -30,8 +30,17 @@ function streamFavicon() {
...
+		$url = wfExpandUrl( $favicon, PROTO_CANONICAL );
+		$client = MediaWikiServices::getInstance()
```
RhinosF1, line 35: Could we use a more specific name here & for $status?

RhinosF1 (changes requested):

modules/mediawiki/files/favicon.php
Universal-Omega: We don't need to. It's already used elsewhere, and this allows the rest to be executed. Just changing here won't work, and I want to maintain consistency.

modules/mediawiki/files/favicon.php
Universal-Omega: Code format. Looks neater to me.

PR #2556, opened 2022-05-15
varnish: force thumb_handler.php to mwtask111

Universal-Omega:
So that image transforms go through mwtask111, and that they don't cause issues on mw*

We may need to give mwtask111 some extra FPM workers if we do this also, I think maybe we should give it 8 (to start with?), setting the multiplier to 4.0 and the minimum to 2 maybe? That should probably be enough for just this. If we do it and it causes issues on mwtask111 something else we may consider is increasing the cores on mwtask111 to handle more of the major loads currently handled solely by mw*. It also may not be necessary to be at 6 on mw* anymore with this no longer there. (provided doing this successfully moves the majority of the converts to mwtask111 that is). Although I do think this by itself may not, I think it would be a start yes, but we may have to experiment further, by changing some things in `$wgLocalFileRepo`, maybe by setting up something allowing us to use `thumbProxyUrl` to proxy to mwtask111, and setting `disableLocalTransform` to `true`? Not entirely sure how it all works, but I do think that at least this PR may be a start. These are just some things we may want to consider in the future for better distribution of resources, and potentially increase the resilience of MediaWiki servers.


RhinosF1:
Note: this is a big change. Up to now, general user traffic has never touched task.

Universal-Omega:
> Note: this is a big change. Up to now, general user traffic has never touched task.

I'm aware of that. But looking at resources on mw*, stuff like transforms really use up a lot of it (it has quite a bit of entries in the PHP slow log also). If we do this, we can also probably up the timeout and memory limit allowed for transforms, allowing them to be more successful, more often, to hopefully fix issues like [T9228](https://phabricator.miraheze.org/T9228) -- and even if it didn't fix that exact issue, it would be beneficial to treat this as "high-intensive", similar to high intensive jobs, and use mwtask111 for them.

Universal-Omega:
I have updated the commit message to provide more thoughts for the future with this also. Some of that will need tasks later, but posting here just to have it down somewhere for now.

paladox:
I think potentially this will increase the load on mwtask111.

RhinosF1:
> I think potentially this will increase the load on mwtask111.

Of course it will. It's a case of whether the impact to task is worse than the gain mw* will get.

Universal-Omega:
> I think potentially this will increase the load on mwtask111.

Yes it will, this is an experiment if it works we were going to consider a request for a server specifically for this if it'll work.

JohnFLewis:
What's the status with this?

PR #2562, opened 2022-05-16
ssl-certificate: remove `--no-verify-ssl` (T8830)

Universal-Omega:


paladox:
Why? I think this will fail because it does the acme challenge over the requested domain (so the https certificate won't be valid).

Universal-Omega:
> Why? I think this will fail because it does the acme challenge over the requested domain (so the https certificate won't be valid).

The mentioned task. If so it can be closed and the task declined.

PR #2564, opened 2022-05-16
varnish: support redirecting paths in SSL redirects.yaml

Universal-Omega:
This is per discussion in the Miraheze meeting yesterday for [T9213](https://phabricator.miraheze.org/T9213), so that we can create a short URL from miraheze.wiki/br to redirect to the Phabricator bug report form, so we can link the short URL in the exception message.

There are other uses for this as well though.

### DIFF
--- a/modules/varnish/templates/mediawiki.conf
+++ b/modules/varnish/templates/mediawiki.conf
@@ -133,11 +133,13 @@ server {
 server {
 	listen 443 ssl http2;
 	listen [::]:443 ssl http2;
-<%- if property['additional_domain'] -%>
+
+	<%- if property['additional_domain'] -%>
 	server_name <%= property['url'] %> <%= property['additional_domain'] %>;
-<%- else -%>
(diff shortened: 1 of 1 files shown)


JohnFLewis:
I feel using the ssl setup for this isn’t appropriate as it would over complicate this unnecessarily. Either a 301 can be used in nginx or perhaps a common set of redirects can be defined through hiera and applied universally instead?

Universal-Omega:
Also there has been some requests for this in the past, which we have declined, but this would add support to do if it is requested again. But we can do whatever is best I guess. Just not 100% sure how to best do it.

JohnFLewis:
I didn’t mean to close this btw

I suppose in those circumstances, it might be better to integrate it this way than manually define them. I thought it was intended for Miraheze.org etc, I didn’t see the .wiki part

paladox:
Looks good to me, merge away!

paladox (approved):

PR #2566, opened 2022-05-17
Add nutcracker monitoring

Universal-Omega:
I will add the docs section for Nutcracker to Tech:Icinga/MediaWiki_Monitoring once this is merged.


modules/role/manifests/mediawiki/nutcracker.pp
```diff
@@ -47,6 +47,27 @@
             override => true,
         }
 
+        monitoring::nrpe { 'nutcracker process':
+            command => '/usr/lib/nagios/plugins/check_procs -c 1:1 -u nutcracker -C nutcracker',
+            docs    => 'https://meta.miraheze.org/wiki/Tech:Icinga/MediaWiki_Monitoring#Nutcracker'
+        }
+
+        monitoring::services { 'nutcracker port 11212':
```
paladox, line 55: These ports aren't open to mon111/icinga. Only the localhost can access them. Requires opening the firewall.

Universal-Omega: So would I add:
```puppet
    $firewall_rules = query_facts('Class[Role::Icinga2]', ['ipaddress', 'ipaddress6'])
    $firewall_rules_mapped = $firewall_rules.map |$key, $value| { "${value['ipaddress']} ${value['ipaddress6']}" }
    $firewall_rules_str = join($firewall_rules_mapped, ' ')
    ferm::service { 'icinga access nutcracker port 11212':
        proto  => 'tcp',
        port   => '11212',
        srange => "(${firewall_rules_str})",
    }

    ferm::service { 'icinga access nutcracker port 11213':
        proto  => 'tcp',
        port   => '11213',
        srange => "(${firewall_rules_str})",
    }
```
?

I was about to add but decided to ask first so if I was wrong wouldn't have to revert.

paladox: yeh, I think so.

Although we listen on localhost so you may not be able to access that port externally:

root@mw101:/home/paladox# netstat -tuln | grep 11212
tcp        0      0 127.0.0.1:11212         0.0.0.0:*               LISTEN

Universal-Omega: I just removed it entirely for now, and just kept process monitoring.

PR #2579, opened 2022-05-21
img-csp: allow all of wikimedia.org

RhinosF1:


modules/varnish/data/csp.yaml
```diff
@@ -69,7 +69,7 @@ img-src:
   - 'data:'
   - '*.miraheze.org'
   - '*.betaheze.org'
-  - 'upload.wikimedia.org'
+  - '*.wikimedia.org'
```
Universal-Omega, line 72:
```suggestion
  - 'upload.wikimedia.org'
  - 'wikimedia.org'
```
this just seems to need `'upload.wikimedia.org'` and `'wikimedia.org'` (the root) according to the CSP warnings for the reasoning for this PR.

Universal-Omega (changes requested):

Universal-Omega (approved):

PR #2588, opened 2022-05-24
Reformat grafana-webhook

Universal-Omega:


modules/monitoring/files/grafana-webhook.py
```diff
@@ -20,11 +20,20 @@ def post():
         with lock:
             lock.acquire()
             try:
-                message = content['state'] + ' : ' + content['title']
-                if content['alerts'][0]['labels']['team'] == 'mediawiki':
-                    message = message + ' https://grafana.miraheze.org/d/GtxbP1Xnk/mediawiki'
                 x = open('/var/log/icinga2/irc.log', 'a+')
```
RhinosF1, line 23: Nit: I probably introduced this but while we are cleaning up can we change x to a nicer name?

modules/monitoring/files/grafana-webhook.py
```diff
@@ -20,11 +20,20 @@ def post():
...
+
+                    page = ''
+                    if alert['labels']['page'] == 'yes':
+                       page = '!sre '
+
+                    message = f'[Grafana] {page}{status}: {description}'
+
+                    if alert['labels']['team'] == 'mediawiki' and not alert['labels']['dashboard']:
+                        message += ' https://grafana.miraheze.org/d/GtxbP1Xnk/mediawiki'
+                    x.write(message)
```
RhinosF1, line 36: Could we check this fits in an irc message?

modules/monitoring/files/grafana-webhook.py
Universal-Omega: Done

modules/monitoring/files/grafana-webhook.py
Universal-Omega: Done. Added some truncation.

modules/monitoring/files/grafana-webhook.py
```diff
@@ -51,7 +51,7 @@ def post():
             finally:
                 lock.release()
                 lock_acquired = True
-    return (message, 200, None)
+    return '', 204
```
RhinosF1, line 54: Thought: could we maybe vary based on if it actually got the end of the try clause? 204 if success and 4/500? if failed?

Universal-Omega: Done.

PR #2592, opened 2022-05-26
add varnish/nginx to puppet111 (adding SSL)

Reception123:
Do not merge! 

This is probably wrong in at least a few ways on the Nginx side so will need review/advice.


modules/letsencrypt/manifests/nginx.pp
```diff
@@ -0,0 +1,16 @@
+# === Class letsencrypt::nginx
+#
+# Nginx config using hiera
+class letsencrypt::nginx {
+    $sslcerts = loadyaml('/etc/puppetlabs/puppet/ssl-cert/certs.yaml')
+    $php_fpm_sock = 'php/fpm-www.sock'
```
Universal-Omega, line 6: Don't need php-fpm I don't think.

hieradata/common.yaml
```diff
@@ -69,6 +68,10 @@ varnish::backends:
     port: 8202
     probe: false
     pool: false
+  puppet111:
+    port: 8205
```
JohnFLewis, line 72: Should be numerical order

modules/varnish/templates/default.vcl
```diff
@@ -165,6 +165,7 @@ sub mw_request {
 		req.http.Host == "sslrequest.miraheze.org"
 	) {
 		set req.backend_hint = mwtask111;
```
Universal-Omega, line 167: The old mwtask111 here should be removed.

modules/letsencrypt/templates/nginx.conf.erb
```diff
@@ -0,0 +1,30 @@
...
+	ssl_certificate /etc/ssl/localcerts/wildcard.miraheze.org-2020-2.crt;
+	ssl_certificate_key /etc/ssl/private/wildcard.miraheze.org-2020-2.key;
+
+	add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload";
+
+	add_header X-Served-By '<%= scope.lookupvar('::fqdn') %>';
+
+	# XSS Protection
+	add_header x-xss-protection "1; mode=block" always;
+Okay
```
JohnFLewis, line 25: Okay?

modules/letsencrypt/templates/nginx.conf.erb
```diff
@@ -0,0 +1,30 @@
+server {
+	# We can only set backlog once per port (so this will be applied to the others below)
+	listen 80 deferred backlog=4096;
+	listen [::]:80 deferred ipv6only=on backlog=4096;
+	listen 443 ssl deferred http2 backlog=4096;
+	listen [::]:443 deferred ssl http2 ipv6only=on backlog=4096;
+
+	server_name miraheze.org www.miraheze.org;
+	root /srv/mediawiki/landing;
```
Universal-Omega, line 9: Make sure landing is be cloned then, and this would also mean /srv/mediawiki would need to exist. Landing is currently done with deploy-mediawiki for MediaWiki, but for this just could be cloned probably.

modules/varnish/templates/default.vcl
```diff
@@ -165,6 +165,7 @@ sub mw_request {
 		req.http.Host == "sslrequest.miraheze.org"
 	) {
 		set req.backend_hint = mwtask111;
+		set req.backend_hint = puppet111;
```
JohnFLewis, line 168: Why do we have it set twice and not once?

JohnFLewis (changes requested):

Universal-Omega (changes requested):

modules/varnish/templates/default.vcl
Reception123: Oh, can it be done with commas? I don't want to remove mwtask111 until the switch is actually done

modules/varnish/templates/default.vcl
Universal-Omega: Well then don't switch it at all, until the switch is to be made I guess.

Reception123:
will add the backend later once I can confirm that things are working fine, so for now that part is commented out.

hieradata/common.yaml
```diff
@@ -69,10 +68,14 @@ varnish::backends:
     port: 8202
     probe: false
     pool: false
-  matomo101:
+  puppet111:
     port: 8203
```
JohnFLewis, line 72: matomo101 is currently 8203 though?

manifests/site.pp
```diff
@@ -100,6 +100,7 @@
     include puppetdb::database
     include role::puppetserver
     include role::salt
+	include letsencrypt::nginx
```
JohnFLewis, line 103: This looks like a tab

hieradata/common.yaml
Reception123: Ah, is what you meant simply that I accidentally put 8205 instead of 8204 and not that I had to change the number to match the order?

manifests/site.pp
Reception123: not anymore :)

Universal-Omega (changes requested):
Don't forget `letsencrypt: true` in puppet111 hieradata either.

modules/letsencrypt/manifests/init.pp
```diff
@@ -84,6 +84,23 @@
         provider  => 'systemd',
         require   => Package['certbot'],
     }
+    
+    git::clone { 'landing':
```
paladox, line 88: Why are we doing this? We should not be loading any real traffic.

modules/letsencrypt/templates/nginx.conf.erb
paladox: Do not load landing on puppet111! Also you may need to copy how we do domains in mw.

modules/varnish/templates/default.vcl
paladox: Shouldn't sslrequest.miraheze.org this be on puppet111 only?

modules/letsencrypt/templates/nginx.conf.erb
Universal-Omega: Yes, I wasn't saying that landing had to be cloned, which seemed to be a bit of a misunderstanding on my part here. I meant that if the web root is set to landing we need landing, but the ideal way would be to set the root to /var/www/challenges or something.

Universal-Omega (changes requested):
Don't forget to move this to puppet-users either:https://github.com/miraheze/puppet/blob/1d68a0e587475acaec1289374f2ac60ec9b7dfe2/modules/users/data/data.yaml#L17

Reception123:
> Don't forget to move this to puppet-users either:
> 
> https://github.com/miraheze/puppet/blob/1d68a0e587475acaec1289374f2ac60ec9b7dfe2/modules/users/data/data.yaml#L17

I was planning to do this part when I also change the scripts for automating private keys, since first I do need to make sure that the new system actually works

Universal-Omega:
> > Don't forget to move this to puppet-users either:
> > https://github.com/miraheze/puppet/blob/1d68a0e587475acaec1289374f2ac60ec9b7dfe2/modules/users/data/data.yaml#L17
> 
> I was planning to do this part when I also change the scripts for automating private keys, since first I do need to make sure that the new system actually works

Well this includes adding the scripts to puppet111, since those are added in the LetsEncrypt module, so not adding it just means puppet-users won't be able to use it on puppet111, until it's added. Up to you how to do it though.

modules/letsencrypt/manifests/nginx.pp
```diff
@@ -0,0 +1,15 @@
...
+# Nginx config using hiera
+class letsencrypt::nginx {
+    $sslcerts = loadyaml('/etc/puppetlabs/puppet/ssl-cert/certs.yaml')
+
+    nginx::site { 'letsencrypt':
+        ensure  => present,
+        content => template('letsencrypt/nginx.conf.erb'),
+    }
+
+    ssl::wildcard { 'mediawiki nginx wildcard': }
```
Universal-Omega, line 12:
```suggestion
    ssl::wildcard { 'letsencrypt nginx wildcard': }
```

Universal-Omega (changes requested):

Universal-Omega (approved):
I can't guarantee it will work, but can't think of anything else right now.

PR #2600, opened 2022-05-28
automate copying private keys for SSL certs (T7582)

Reception123:


Universal-Omega (approved):

JohnFLewis:
You shouldn’t be using the master source repo to push changes and stage commits

Reception123:
> You shouldn’t be using the master source repo to push changes and stage commits

Should I be using /home/puppet-users/ssl-keys then? Or is there another repo? (I've always used the puppet-users dir out of habit so I wouldn't know)

JohnFLewis:
The puppet-users one is the only copy I know of currently, unless someone has cloned it elsewhere as well

Reception123:
That's fine then, I actually thought if I did that someone would say I should do it directly, but I guess it's the opposite :P

PR #2614, opened 2022-05-30
Grant puppet-users access to ssl-acme

Universal-Omega:


Reception123 (approved):

Reception123:
in case something happens with renewals, it's good to allow puppet-users to manually generate rather than them having to go through the whole process and also having to push to GH

PR #2616, opened 2022-05-30
tests: assert no double space in command

RhinosF1:


modules/mediawiki/files/bin/deploy-mediawiki.py
```diff
@@ -151,15 +151,14 @@ def _construct_rsync_command(time: str, dest: str, recursive: bool = True, local
...
-    else:
-        extrap = ''
+        extrap += '--recurse-submodules '
 
     if branch:
-        extrap += f' origin {branch}'
+        extrap += f'origin {branch} '
 
-    return f'sudo -u {DEPLOYUSER} git -C {_get_staging_path(repo)} pull {extrap} --quiet'
+    return f'sudo -u {DEPLOYUSER} git -C {_get_staging_path(repo)} pull{extrap}--quiet'
```
paladox, line 161: This looks wrong. If you look above there’s no space before so this would make it pull—recurse-submodules

paladox: Ohhhh, you added a space at the start variable.

RhinosF1: No it wouldn't. Please read the tests that prove the output and how it is constructed.

PR #2622, opened 2022-05-31
letsencrypt: try suppressing InsecureRequestWarning using env

MacFan4000:
other method had no effect


Universal-Omega:
CI failing.

PR #2629, opened 2022-05-31
Attempt to fix grafana-webhook

Universal-Omega:


modules/monitoring/files/grafana-webhook.py
```diff
@@ -21,22 +21,23 @@ def post():
         with lock:
             lock.acquire()
             try:
+                alerts = content['alerts']
                 messages = []
-                for alert in content['alerts']:
-                    status = alert['status']
-                    summary = alert['annotations']['summary']
+                for i in alerts:
```
RhinosF1, line 26: Please don't use single letter variable names.

Universal-Omega: "i" is a fairly standard and simple variable name, so it's why I did here. It keeps the rest of the code simple by having just a one letter index, as it's necessary numerous times. If not this, what would you recommend?

PR #2644, opened 2022-06-02
archive puppet-users, create letsencrypt-admins group

Reception123:
Per T9290#188843, since the group now has sudo access and it doesn't make sense to use the word puppet considering it has nothing to do with puppet, this makes more sense

### DIFF
--- a/modules/puppetserver/manifests/init.pp
+++ b/modules/puppetserver/manifests/init.pp
@@ -140,10 +140,10 @@ class puppetserver(
         ],
     }
 
-    file { '/home/puppet-users':
+    file { '/home/letsencrypt-admins':
         ensure  => directory,
         owner   => 'root',
-        group   => 'puppet-users',
+        group   => 'letsencrypt-admins',
         mode    => '0770',
     }
  
--- a/modules/users/data/data.yaml
+++ b/modules/users/data/data.yaml
@@ -34,6 +34,7 @@ groups:
     description: (archived group) limited access on puppet111 to manage SSL certificates
     members: []
     privileges: []
+
   bastion:
     gid: 2005
     description: users who require bastion access
@@ -43,7 +44,7 @@ groups:
   letsencrypt-admins:
     gid: 2006
     description: limited sudo access on puppet111 to manage SSL certificates
-    members: []
+    members: [macfan]
     privileges: ['ALL = (ALL) NOPASSWD: /root/ssl-certificate',
                  'ALL = (ALL) NOPASSWD: /var/lib/nagios/ssl-acme']
 
--- a/modules/users/data/data.yaml
+++ b/modules/users/data/data.yaml
@@ -31,7 +31,7 @@ groups:
                  'ALL = (ALL) NOPASSWD: /bin/journalctl *']
   puppet-users:
     gid: 2004
-    description: (archived group) limited access on puppet111 to manage SSL certificates
+    description: limited access on puppet servers
     members: []
     privileges: []
 
@@ -43,7 +43,7 @@ groups:
     
   letsencrypt-admins:
     gid: 2006
-    description: limited sudo access on puppet111 to manage SSL certificates
+    description: limited sudo access to manage SSL certificates
     members: [macfan]
     privileges: ['ALL = (ALL) NOPASSWD: /root/ssl-certificate',
                  'ALL = (ALL) NOPASSWD: /var/lib/nagios/ssl-acme']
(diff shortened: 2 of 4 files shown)


Universal-Omega:
Just now thinking, but is letsencrypt-admins even accurate? Non-LetsEncrypt certificates can be done also, so shouldn't it be ssl-admins or something?

Universal-Omega:
Also, don't forget to rename the monitoring group (puppetusers) as well. That should be able to just be renamed unlike the access group, as we have renamed monitoring groups before.

MacFan4000:
just a reminder to also update directory permissions so that we can debug the script or add non LE private keys when needed. (That includes both /etc/puppetlabs/puppet/ssl-keys and /home/puppet-users/ssl-keys)

modules/users/data/data.yaml
```diff
@@ -31,15 +31,22 @@ groups:
...
+
   bastion:
     gid: 2005
     description: users who require bastion access
     members: [macfan, rhinos, universalomega, agent]
     privileges: []
+    
+  letsencrypt-admins:
+    gid: 2006
+    description: limited sudo access on puppet111 to manage SSL certificates
```
Universal-Omega, line 46:
```suggestion
    description: limited sudo access to manage SSL certificates
```
Dropping puppet111, as it's one less thing to have to change if it's ever changed again. But it's up to you.

modules/users/data/data.yaml
```diff
@@ -31,15 +31,22 @@ groups:
                  'ALL = (ALL) NOPASSWD: /bin/journalctl *']
   puppet-users:
     gid: 2004
-    description: limited access on puppet111 to manage SSL certificates
-    members: [macfan]
-    privileges: ['ALL = (ALL) NOPASSWD: /root/ssl-certificate',
-                 'ALL = (ALL) NOPASSWD: /var/lib/nagios/ssl-acme']
+    description: (archived group) limited access on puppet111 to manage SSL certificates
```
Universal-Omega, line 34:
```suggestion
    description: (archived group)
```
I think that this can just say archived group, that way it doesn't mention puppet111, so it doesn't have to be changed if that's ever changed, and the rest of the description is unnecessary if it's archived. Up to you though.

Universal-Omega (changes requested):

Reception123:
> Just now thinking, but is letsencrypt-admins even accurate? Non-LetsEncrypt certificates can be done also, so shouldn't it be ssl-admins or something?

That's what I thought initially, but the reason for why I chose letsencrypt-admins is because the directory/class is called letsencrypt so I think we should be consistent. Though I'm not too sure why that's called that either..

Universal-Omega:
> > Just now thinking, but is letsencrypt-admins even accurate? Non-LetsEncrypt certificates can be done also, so shouldn't it be ssl-admins or something?
> 
> That's what I thought initially, but the reason for why I chose letsencrypt-admins is because the directory/class is called letsencrypt so I think we should be consistent. Though I'm not too sure why that's called that either..

I'm only bringing it up as once we create the new group, we can't change the name again later (easily) I can do a separate PR for this and renaming the entire LetsEncrypt module if you'd rather as well?

modules/puppetserver/manifests/init.pp
```diff
@@ -140,10 +140,10 @@
         ],
     }
 
-    file { '/home/puppet-users':
+    file { '/home/letsencrypt-admins':
```
Universal-Omega, line 143: We can rather absent this, and do the ensure for the letsencrypt-admins home directory in the LetsEncrypt module, since it is for that, if this were to ever change also (although highly unlikely) it would need moved anyways, so for consistency maybe this should be in the LetsEncrypt module?

Universal-Omega (changes requested):
Additionally, still need the monitoring group renamed as well.
* https://github.com/miraheze/puppet/blob/f507b96989db0a60797804ee2fa8906ab5091878/modules/monitoring/files/hosts.conf#L6-L7
* https://github.com/miraheze/puppet/blob/1d68a0e587475acaec1289374f2ac60ec9b7dfe2/modules/monitoring/files/users.conf#L75
* https://github.com/miraheze/puppet/blob/1d68a0e587475acaec1289374f2ac60ec9b7dfe2/modules/monitoring/files/users.conf#L115-L117

JohnFLewis:
@Reception123 status?

Universal-Omega:
> @Reception123 status?

This is superceded by #2657 I think.

Reception123:
Closing per above, a better solution (merging) has been proposed

PR #2645, opened 2022-06-02
ssl-acme: skip checking the host key

MacFan4000:
if we don't skip it then the auto-renewal script fails

### DIFF
--- a/modules/letsencrypt/files/ssl-acme
+++ b/modules/letsencrypt/files/ssl-acme
@@ -58,7 +58,7 @@ WARNING)
         ;;
         HARD)
                 # Lets Renew here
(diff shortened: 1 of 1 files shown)


RhinosF1 (changes requested):
You need to find where to place the key and add it.

MacFan4000:
> You need to find where to place the key and add it.

If you mean add it server side (not in puppet) it’s not possible for me to do that as my access is very limited.

This is a simple solution that skips the check just for this script.

RhinosF1:
This might be simple but it's also dangerous and poses security implications. Not really bothered whether it's done in puppet or on server.

Universal-Omega:
This has security risk to do also.

PR #2648, opened 2022-06-04
make betaheze.org use landing

Universal-Omega:


Universal-Omega:
Just realized this won't work due to SSL certs.

PR #2657, opened 2022-06-06
Merge letsencrypt and ssl modules; create ssl-admins group

Universal-Omega:
Per [T9347](https://phabricator.miraheze.org/T9347).


JohnFLewis:
It feels weird we're renaming letsencrypt to ssl because it "does more SSL than just letsencrypt" but are renaming a module that is SSL to something else? It sounds like these two modules could just be merged

Universal-Omega:
> It feels weird we're renaming letsencrypt to ssl because it "does more SSL than just letsencrypt" but are renaming a module that is SSL to something else? It sounds like these two modules could just be merged

Probably could, and I even considered that with this, but the fact that the SSL module itself is just SSL certs, I figured it could be called just that, as to me it's even a more representive name than just SSL for what it actually does. However, I can do that if you want. The difference between three two tho is that ssl (now sslcert with this) is certificates applied to all servers, while letsencrypt (now ssl with this) is only applied to a specific server, so I wasn't sure if merging them was the correct approach. However, if we really just want to keep it the way it is and ignore the fact that letsencrypt does more than just letsencrypt certs, and just leave it as that, than this PR can really just be closed.

JohnFLewis:
I think they should both be merged - one part handles the automation and deployment of SSL certs on all servers, the other handles the adding and processing of them to the SSL repo. One's a pre and ones a post process, but they're part of the same chain.

Universal-Omega:
> I think they should both be merged - one part handles the automation and deployment of SSL certs on all servers, the other handles the adding and processing of them to the SSL repo. One's a pre and ones a post process, but they're part of the same chain.

Done in this PR now.

JohnFLewis:
It looks okay to me

Reception123 (approved):

PR #2659, opened 2022-06-08
Move `mediawiki::jobqueue::chron` include to redis role

Universal-Omega:
Also changes the redis role description since we don't use redis for caching anymore, only for jobchron.

### DIFF
--- a/manifests/site.pp
+++ b/manifests/site.pp
@@ -38,7 +38,6 @@ node 'graylog121.miraheze.org' {
 node 'jobchron121.miraheze.org' {
     include base
     include role::redis
-    include mediawiki::jobqueue::chron
 }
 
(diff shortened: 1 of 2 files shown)


JohnFLewis:
Also we use redis for matomo101 btw

Universal-Omega:
> Also we use redis for matomo101 btw

I just realised that. Matomo doesn't use the redis role but regardless, this isn't worth it, as we may use this role for something else again in the future.

Universal-Omega:
> Can we rename the role as it no longer appears this is a redis only role if it’s including MediaWiki related things
> 
> tbh MediaWiki::jobqueue::chron seems more descriptive, accurate and consistent than using a role - so perhaps we should just merge into the redis role to that?

Oh yeah, we could probably do that also.

PR #2672, opened 2022-06-15
matomo: Add missing grants to piwik grants (matomo)

paladox:


paladox:
From https://matomo.org/faq/troubleshooting/how-do-i-check-if-my-mysql-user-has-all-required-grants/

PR #2683, opened 2022-06-24
mwscript: add beta database list support

Universal-Omega:
Also removes deleted, since that isn't a combi list, running foreachwikiindblist actually doesn't work.


RhinosF1 (commented):
Replaced by #2684

PR #2687, opened 2022-07-01
rhinosf1: remove access to mw cluster

RhinosF1:
I'm still staying around but I'm fairly busy.

I've spoke to @Reception123 and given my focus is mainly the devops stuff with deploy tool and our other python scripts. I can maintain them without shell.


Reception123:
Sorry to see you go and thanks for offering to stay around.

PR #2703, opened 2022-07-25
varnish: force thumb_handler through a specific set of servers

paladox:
Also remove mw102 and mw122 (these will be used for thumb_handler).


Universal-Omega:
This can probably be closed for now.

PR #2706, opened 2022-07-26
mariadb: tweak config

paladox:
* Changes buffer pool to be 80% of ram available to db*.


Universal-Omega:
Just wondering what the status of this is, and also the purpose/need of it, out of my own curiosity?

Universal-Omega:
Also, has conflict.

PR #2713, opened 2022-07-27
users: add varnishlog and varnishstat to cache-admins

Universal-Omega:
Also fix description key.


Reception123:
@JohnFLewis Any objections to this? I think it makes sense to allow cache-admins to have access to that kind of information to assist them with investigating.

Reception123:
Approved in https://phabricator.miraheze.org/T9598

PR #2767, opened 2022-08-07
varnish: add index.php to vcl_hash

Universal-Omega:
`/w/index.php?title=Main_Page`
> Age 0
> X-Cache cp30 PASS UNCACHEABLE

`/wiki/Main_Page`
> Age 7785
> X-Cache cp30 HIT (6)

And this was the only thing I could really see as differing here. So I thought it may have something to do with the cache splitting here.

### DIFF
--- a/modules/varnish/templates/default.vcl
+++ b/modules/varnish/templates/default.vcl
@@ -322,7 +322,7 @@ sub vcl_recv {
 # Defines the uniqueness of a request
 sub vcl_hash {
 	# FIXME: try if we can make this ^/wiki/ only?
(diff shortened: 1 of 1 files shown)


JohnFLewis:
This is uncachable as the CACHE-CONTROL header marks this as a private response. Therefore this is a HFP here https://github.com/miraheze/puppet/blob/master/modules/varnish/templates/default.vcl#L352

Universal-Omega:
> This is uncachable as the CACHE-CONTROL header marks this as a private response. Therefore this is a HFP here https://github.com/miraheze/puppet/blob/master/modules/varnish/templates/default.vcl#L352

Makes sense. Thanks!

PR #2772, opened 2022-08-10
Raise varnish::interval-check to 60 seconds again

Universal-Omega:
With it low (at 20 seconds), 503s would be even more common, I think. So this reverts it back to 60 seconds now that gluster migration is done.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -18,7 +18,7 @@ memcached_servers_1:
   - 2a10:6740::6:105:11211:1
 memcached_servers_3:
   - 2a10:6740::6:405:11211:1
-varnish::interval-check: '20s'
+varnish::interval-check: '60s'
 varnish::interval-timeout: '60s'
 varnish::backends:
   mw121:


JohnFLewis:
Having this low increases the time to failure and time to recover. With this at 60, it would take 2 minutes before a failure is detected, with it at 20, it would take 40s. Increasing this would mean we’re pushing more and more traffic to failed servers, meaning we’d be down for longer in the long run

Universal-Omega:
> Having this low increases the time to failure and time to recover. With this at 60, it would take 2 minutes before a failure is detected, with it at 20, it would take 40s. Increasing this would mean we’re pushing more and more traffic to failed servers, meaning we’d be down for longer in the long run

That is true yeah, which I am guessing why this was even done, but it also increases the amount of temporary 503s. But I am fine with leaving, I only did this PR since I thought it was lowered just for the gluster migration higher load. Feel free to close this if you want to leave it.

JohnFLewis:
I’d rather have temporary 503s that get fixed quickly, than loads of request hanging and longer 503s that are permanent with more failed requests inbetween

Universal-Omega:
> I’d rather have temporary 503s that get fixed quickly, than loads of request hanging and longer 503s that are permanent with more failed requests inbetween

Makes sense. Thanks!

JohnFLewis:
On second thought, I will merge this for now as the timeout is still 60s, this reverts back to its original state and a more thorough review of Varnish tuning will need to be done once the infra changes are done and a potential cp40 is introduced.

PR #2775, opened 2022-08-10
check_reverse_dns: allow cname flattening for porkbun dns

MacFan4000:


modules/monitoring/files/check_reverse_dns.py
```diff
@@ -78,7 +78,7 @@ def check_records(hostname):
         for nameserver in nameserversans:
             nameserver = str(nameserver)
             nameservers.append(nameserver)
-            if nameserver.endswith('.ns.cloudflare.com.') or nameserver.endswith('.dreamhost.com.'):
+            if nameserver.endswith('.ns.cloudflare.com.') or nameserver.endswith('.dreamhost.com.') or nameserver.endswith('.ns.porkbun.com.'):
```
paladox, line 81: I think that a config should be used to specify this but that can be done in a separate commit.

PR #2777, opened 2022-08-11
check_rdns: update cf_at_root logic

RhinosF1:
use a cleaner list and replace some Cloudflare specific naming with generic stuff.


RhinosF1:
@paladox: as requested.

modules/monitoring/files/check_reverse_dns.py
```diff
@@ -78,8 +78,14 @@ def check_records(hostname):
         for nameserver in nameserversans:
             nameserver = str(nameserver)
             nameservers.append(nameserver)
-            if nameserver.endswith('.ns.cloudflare.com.') or nameserver.endswith('.dreamhost.com.') or nameserver.endswith('.ns.porkbun.com.'):
-                uses_cf_at_root = True
+            flatten_manadatory_providers = [
+                'ns.cloudlfare.com',
+                'dreamhost.com',
+                'porkbun.com',
```
Universal-Omega, line 84:
```suggestion
                'ns.porkbun.com',
```
For what it was before anyway.

modules/monitoring/files/check_reverse_dns.py
```diff
@@ -78,8 +78,14 @@ def check_records(hostname):
...
+            ]
+            for provider in flatten_manadatory_providers:
+                if nameserver.endswith(f'.{provider}'):
```
Universal-Omega, line 87:
```suggestion
                if nameserver.endswith(f'.{provider}.'):
```

Universal-Omega: Also, endswith supports a tuple I think, so if list above is converted to a tuple, we do not need the loop I think?

https://docs.python.org/3/library/stdtypes.html#str.endswith

> suffix can also be a tuple of suffixes to look for.

Universal-Omega (changes requested):

modules/monitoring/files/check_reverse_dns.py
RhinosF1: No reason why I can't make it a tuple

Universal-Omega: This should be enough here also, to just add minor simplification I think, but that is up to you.
```lang=py
if nameserver.endswith(('.ns.cloudlfare.com.', '.dreamhost.com.', '.ns.porkbun.com.')):
    cname_check_impossible = True
```

RhinosF1: It's not simpler imho. Moving it to a tuple outside with each entry on a new line looks cleaner.

Universal-Omega: Feel free to do however you feel best then.

Universal-Omega (approved):

PR #2796, opened 2022-08-27
use ldap cname instead of ldap server directly

Universal-Omega:


paladox:
The ability to change the ldap host should stay.

Universal-Omega:
> The ability to change the ldap host should stay.

well it is changeable, in DNS, this just makes it only have to be changed in one place, no?

Universal-Omega:
Merging per conversation with paladox on Discord.

PR #2858, opened 2022-09-23
swift: allow Icinga2 access to firewall

Universal-Omega:


modules/role/manifests/swift.pp
```diff
@@ -21,7 +21,7 @@
         include role::memcached
 
         $firewall_rules_mediawiki_str = join(
-            query_facts('Class[Role::Mediawiki]', ['ipaddress', 'ipaddress6'])
```
paladox, line 24: Can remove this and just use firewall_rules_swift_str

Universal-Omega: Done

PR #3040, opened 2022-12-01
base::monitoring: lower critical/warning threshold for Disk Space

paladox:


JohnFLewis:
10 and 5 are low enough in my opinion as when we have a 20GB server - this lowers warn down to 1GB

PR #3043, opened 2022-12-03
redis: disable lua-time-limit

paladox:
Either it's a coincidence that issues with the job queue stopped after changing this or this fixed the issue.

Either way I don't see why we need this set at 5s. It was imported 8 years ago.

This defaults it to the default in redis which is 500ms.


paladox:
oh it is a conincidence. I misread the default, it's 5s.

PR #3054, opened 2022-12-06
mediawiki: add tempurl to PrivateSettings.php

Universal-Omega:


modules/swift/templates/proxy-server.conf.erb
```diff
@@ -21,6 +21,10 @@ account_autocreate = true
 conn_timeout = 2.0
 node_timeout = 20
 
+[filter:tempurl]
+use = egg:swift#tempurl
```
paladox, line 25: This is already defined. Line 55.

Universal-Omega: Oh. Thanks.

PR #3073, opened 2022-12-15
ssl-acme: simplify now that ssl-certificate auto-pushes keys on renewal

MacFan4000:


Universal-Omega:
Doesn't ssl-certificate still need updated to IE fix commit message to be Update cert for renewals?

MacFan4000:
> Doesn't ssl-certificate still need updated to IE fix commit message to be Update cert for renewals?

Ssl-certificate uses the same commit message  for renewals that ssl-acme did

PR #3074, opened 2022-12-15
ssl-certificates: for generation handle certbot errors and don't auto…

MacFan4000:
…-add public keys for wildcard certs

generally wildcard certs won't be implemented quite the same as normal ones

the error handling prevents erroneous entries in certs.yaml when certbot had failed but the script kept going

### DIFF
--- a/modules/ssl/files/ssl-certificate.py
+++ b/modules/ssl/files/ssl-certificate.py
@@ -110,11 +110,11 @@ class SslCertificate:
             if self.no_existing_key:
                 out = os.system(f'/usr/bin/certbot --force-renewal --expand --no-verify-ssl certonly --manual --preferred-challenges dns-01 {self.overwrite} -d {self.domain} {self.secondary_domain}')
                 if out != 0:
-                     sys.exit("Error: Certbot failed (either the domain isn't pointed or we are being rate limited)")
+                    sys.exit("Error: Certbot failed (either the domain isn't pointed or we are being rate limited)")
             else:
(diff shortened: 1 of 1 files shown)


modules/ssl/files/ssl-certificate.py
```diff
@@ -107,9 +108,13 @@ def generate_letsencrypt_certificate(self):
                 print('Generating Wildcard SSL certificate with LetsEncrypt')
 
             if self.no_existing_key:
-                os.system(f'/usr/bin/certbot --force-renewal --expand --no-verify-ssl certonly --manual --preferred-challenges dns-01 {self.overwrite} -d {self.domain} {self.secondary_domain}')
+                out = os.system(f'/usr/bin/certbot --force-renewal --expand --no-verify-ssl certonly --manual --preferred-challenges dns-01 {self.overwrite} -d {self.domain} {self.secondary_domain}')
+                if out != 0:
+                     sys.exit("Error: Certbot failed (either the domain isn't pointed or we are being rate limited)")
```
Universal-Omega, line 113: We need to know the exact rate limit error. Different errors from this is different lengths for the rate limit, to know how long the rate limit lasts, we don't a generic catch-all error.
Including below also.

Universal-Omega (changes requested):

modules/ssl/files/ssl-certificate.py
MacFan4000: assuming -q wasn't used we would still see the output from certbot

Universal-Omega (approved):

PR #3113, opened 2023-01-06
postfix: discard mail with spamassassin level of 7 or higher

Universal-Omega:


modules/postfix/files/main.cf
```diff
@@ -46,3 +46,5 @@ notify_classes = delay, resource, software
 smtpd_sender_login_maps = ldap:/etc/postfix/ldap/smtpd_sender_login_maps
 smtpd_sender_restrictions = reject_sender_login_mismatch
 strict_mailbox_ownership = no
+
+header_checks = regexp:/^X-Spam-Level: [^*]*[*]{7,}$/ DISCARD
```
JohnFLewis, line 50: For consistency, this should be done in a separate file similar to the LDAP protocol ones above rather than directly in the file.

JohnFLewis (changes requested):

modules/postfix/files/main.cf
Universal-Omega: Sure, I'll make the change later. I thought about that also, but decided this way as if avoided a second file, but I understand the need for consistency here, so that is fine also. I'll make the change later. Thanks!

Universal-Omega: Done.

Reception123:
@JohnFLewis Could you review this when you have the time?

PR #3114, opened 2023-01-06
explicitly configure phabricator favicon

The-Voidwalker:
Config documentation is limited to https://secure.phabricator.com/T13103#237103


The-Voidwalker:
Not necessary (unless we want to use a custom favicon)

PR #3118, opened 2023-01-11
prevent mw-admins from using swift --delete all

Reception123:
There would never be any need to use it, inherently dangerous command


JohnFLewis:
Blacklisting in sudo is extremely ineffective - that's why its traditionally designed as a whitelist. In theory, someone could create a file, put that command in it,  run it, and sudoers will allow it as running the file isn't blacklisted.

PR #3119, opened 2023-01-11
allow ssl-admins to run puppet on ns* via Salt for T7582

Reception123:
As a disclaimer this could either be completely wrong, not work or not be the correct way to do it but I thought I might as well give it a try.


Reception123:
not a valid way to proceed

PR #3121, opened 2023-01-14
T10285: Set Content-Disposition header on static.miraheze.org requests containing ?download

redbluegreenhat:
This pull request sets Content-Disposition: attachment on requests at static.miraheze.org that contain ?download.

This is required by Extension:MultiMediaViewer in order to actually download the file: see https://phabricator.miraheze.org/T10285.

The code is from Gilles <gdubuc@wikimedia.org> (https://gerrit.wikimedia.org/r/c/operations/puppet/+/120617), I just ported it to Miraheze.

### DIFF
--- a/modules/varnish/templates/default.vcl
+++ b/modules/varnish/templates/default.vcl
@@ -197,6 +197,14 @@ sub mw_request {
 		if (req.method == "OPTIONS" && req.http.Origin) {
 			return (synth(200, "CORS Preflight"));
 		}
+		# From Wikimedia: https://gerrit.wikimedia.org/r/c/operations/puppet/+/120617/7/templates/varnish/upload-frontend.inc.vcl.erb
+		# required for Extension:MultiMediaViewer: T10285
+		if (req.url ~ "(?i)(\?|&)download(=|&|$)") {
+			/* Pretend that the parameter wasn't there for caching purposes */
+			set req.url = regsub(req.url, "(?i)(\?|&)download(=[^&]+)?$", "");
+			set req.url = regsub(req.url, "(?i)(\?|&)download(=[^&]+)?&", "\1");
+			set req.http.X-Content-Disposition = "attachment";
+		}
 	}
 
 	# Don't cache a non-GET or HEAD request
--- a/modules/varnish/templates/default.vcl
+++ b/modules/varnish/templates/default.vcl
@@ -203,7 +203,7 @@ sub mw_request {
 			/* Pretend that the parameter wasn't there for caching purposes */
 			set req.url = regsub(req.url, "(?i)(\?|&)download(=[^&]+)?$", "");
 			set req.url = regsub(req.url, "(?i)(\?|&)download(=[^&]+)?&", "\1");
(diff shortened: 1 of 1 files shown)


redbluegreenhat:
As for the copyright status of this port, I believe Gilles' original code is trivial enough to not warrant copyright protection. 
The initial regex that looks for the download GET parameter contains simply a description of GET parameters (look for either ? (first parameter) or & (every parameter after the first), then the string download, then either = (parameter value) or & (next parameter)). Trying to do this with something like PHP's str_contains, that looks only if the substring is in the string is not useful for this use case, as it could lead to unwanted matches on files with download in their names, or wikis with the database name downloadwiki or download(whatever)wiki. There's really no other way to do this efficiently.

The following two statements remove the download parameter (I think the second statement never executes in most circumstances, as at that point the download parameter has been removed by the first statement, could run if there's a second download GET parameter.), then sets the Content-Disposition parameter in the only way you can set it on varnish. Literally no other way to do this in varnish.

Similarly, this port to Miraheze may not warrant copyright protection. However, in case my modifications made this eligible for copyright, it can be used and distributed under the GNU Project's GPL-3.0, the license specified at this repo's LICENSE file as of writing.

redbluegreenhat:
I think I just did a big oops while trying to rewrite the history of my local OAuth2 fork

modules/varnish/templates/default.vcl
```diff
@@ -197,6 +197,14 @@ sub mw_request {
...
 			return (synth(200, "CORS Preflight"));
 		}
+		# From Wikimedia: https://gerrit.wikimedia.org/r/c/operations/puppet/+/120617/7/templates/varnish/upload-frontend.inc.vcl.erb
+		# required for Extension:MultiMediaViewer: T10285
+		if (req.url ~ "(?i)(\?|&)download(=|&|$)") {
+			/* Pretend that the parameter wasn't there for caching purposes */
+			set req.url = regsub(req.url, "(?i)(\?|&)download(=[^&]+)?$", "");
+			set req.url = regsub(req.url, "(?i)(\?|&)download(=[^&]+)?&", "\1");
+			set req.http.Content-Disposition = "attachment";
+		}
```
paladox, line 207: Looks like you need to add:

```
if (req.http.X-Content-Disposition == "attachment") {
  set resp.http.Content-Disposition = "attachment";
}
```

somewhere from line 508?

redbluegreenhat: This is skipped in this version, while this is done in the Wikimedia version, this one just sets Content-Disposition directly. To be honest, I don't understand why this is done that way in Wikimedia, I think I'm just not understanding something about Varnish.

paladox: Well this sends it to the backend the other one is responding to the client I think.

redbluegreenhat: > Well this sends it to the backend the other one is responding to the client I think.

Makes sense. I'll add the "if" to vcl_deliver.

PR #3123, opened 2023-01-15
Do not cache OAuth2 requests

redbluegreenhat:
Currently, Varnish doesn't cache OAuth1 requests, via looking for the OAuth string in the Authorization header. This only matches OAuth1 requests, not OAuth2 requests.

This PR fixes that by looking for the "Bearer" string in the Authorization header, which is included in OAuth2 requests.

For reference, OAuth2's Authorization header looks like this: Authorization: Bearer (OAuth2 token)

### DIFF
--- a/modules/varnish/templates/default.vcl
+++ b/modules/varnish/templates/default.vcl
@@ -234,7 +234,7 @@ sub mw_request {
 	}
 
 	# A requet via OAuth should not be cached or use a cached response elsewhere
-	if (req.http.Authorization ~ "OAuth") {
+	if (req.http.Authorization ~ "OAuth" || req.http.Authorization ~ "Bearer") {
 		return (pass);
 	}
 


redbluegreenhat:
I think this may not be necessary after all. While OAuth 1 uses dedicated special pages for the user authentication grant, OAuth2 works over rest.php for user authentication, which should already be prevented from caching thanks to https://github.com/miraheze/puppet/blob/db89e193437bf84a3c9ca953319239f834688453/modules/varnish/templates/default.vcl#L232.

RhinosF1:
@redbluegreenhat: OAuth2 uses rest.php in all cases I’m aware of. Please also refrain from discussing security issues in public. They should be reported in Phabricator where you will be given access to send private patches via the github security reporting scheme.

redbluegreenhat:
> @redbluegreenhat: OAuth2 uses rest.php in all cases I’m aware of. Please also refrain from discussing security issues in public. They should be reported in Phabricator where you will be given access to send private patches via the github security reporting scheme.

Yeah, I'm very sorry about that, I didn't realize that this could be a security issue until I opened this PR. I'll make sure to report this through the correct channels in the future if I ever find something like this

PR #3146, opened 2023-02-08
renewal - don't use -q

MacFan4000:
this makes issues easier to debug, is no longer needed since the pushing is no longer done in this script


Universal-Omega:
Wait, doesn't this set newprivate also, do we want renewals to set that?

PR #3155, opened 2023-02-13
Implement deleting backups within miraheze-backup.py

paladox:


JohnFLewis:
Why do we need to delete backups manually? Their cost is insignificant and they get auto-removed after a month.

modules/base/templates/backups/miraheze-backup.py.erb
```diff
@@ -14,10 +14,12 @@ from fabric import Connection
 from datetime import datetime
 
 parser = argparse.ArgumentParser(description='Provides backup and download functionality for Miraheze backups from PCA.')
-parser.add_argument('action', choices=['backup', 'download', 'find', 'unfreeze'], help='Action to be ran')
+parser.add_argument('action', choices=['backup', 'download', 'find', 'delete', 'unfreeze'], help='Action to be ran')
 parser.add_argument('type', choices=['private', 'sslkeys', 'phabricator', 'sql', 'mediawiki-xml'], help='Type of backup to handle using the action')
 parser.add_argument('--date', dest='date', help='Date for backup to download', metavar='YYYY-MM-DD')
 parser.add_argument('--database', dest='database', help='Specific database to download or backup')
+parser.add_argument('--delete-all', dest='delete_all', required=False, action='store_true', default=False, help='Delete all backups for the "type" selected')
+parser.add_argument('--file-name', dest='file_name', help='Specify the file to delete. To be used with "delete" action')
```
JohnFLewis, line 22: This parameter goes against a structured system - if you call the type of backup and the date, why do you need to specify a filename when you can construct it from both of those facts?

modules/base/templates/backups/miraheze-backup.py.erb
```diff
@@ -14,10 +14,12 @@ from fabric import Connection
 from datetime import datetime
 
 parser = argparse.ArgumentParser(description='Provides backup and download functionality for Miraheze backups from PCA.')
-parser.add_argument('action', choices=['backup', 'download', 'find', 'unfreeze'], help='Action to be ran')
+parser.add_argument('action', choices=['backup', 'download', 'find', 'delete', 'unfreeze'], help='Action to be ran')
 parser.add_argument('type', choices=['private', 'sslkeys', 'phabricator', 'sql', 'mediawiki-xml'], help='Type of backup to handle using the action')
 parser.add_argument('--date', dest='date', help='Date for backup to download', metavar='YYYY-MM-DD')
 parser.add_argument('--database', dest='database', help='Specific database to download or backup')
+parser.add_argument('--delete-all', dest='delete_all', required=False, action='store_true', default=False, help='Delete all backups for the "type" selected')
```
JohnFLewis, line 21: This would also delete every backup - meaning you lose all historic archives that you might want to intentionally keep - therefore this feature seems really dangerous to me if you don't restrict it by date.

paladox:
> Why do we need to delete backups manually? Their cost is insignificant and they get auto-removed after a month.

Oh, didn't realise they auto deleted.

PR #3176, opened 2023-04-01
[DNM] bookworm/python: use bookworm-only features, drop buster dependencies

RhinosF1:


RhinosF1:
Please ping me before this gets rolled out whenever we are on bookworm

Universal-Omega:
Per above

PR #3203, opened 2023-04-19
Update my SSH key

redbluegreenhat:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

I've updated my SSH key to a FIDO2 one. For extra assurance that it is me, this commit is signed using the SSH key I've used in almost all my commits in Github, and this commit message is also signed using my OpenPGP key, which I've used in almost all my emails to sre@miraheze.org
-----BEGIN PGP SIGNATURE-----

iHUEARYIAB0WIQQRreQ5NgDBvf/LwKWY3hWUKwjKAAUCZEAg4AAKCRCY3hWUKwjK
APhkAP9ot1Awbosqb976m7geD2FF25Kv2TsXze9W7X0qb+V1VAEAp0XsqtNRSiTs
0hGf4awYHOkEmt3VAE7TazpH3iwkuw4=
=iYK3
-----END PGP SIGNATURE-----

### DIFF
--- a/modules/users/data/data.yaml
+++ b/modules/users/data/data.yaml
@@ -151,5 +151,5 @@ users:
     ensure: present
     uid: 1016
     name: alex
-    realname: Alex 
-    ssh_keys: [ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAICPdPDIUcsM/V6/uypn/mAlVB4gPnpF5lHrdLSqQowEr alex@blueselene.com]
+    realname: Alex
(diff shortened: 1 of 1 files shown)


Reception123:
Due to the requirement set out in https://meta.miraheze.org/wiki/Tech:Appointment_and_revocation_policy#New_Access I would like to confirm with @The-Voidwalker whether we can allow this exception to the rule due to my limited knowledge in this area.

Reception123:
In that case if it is an ed25519 there is no problem, will merge.

PR #3210, opened 2023-04-26
nginx: remove use of ipv6only

The-Voidwalker:
this option is enabled by default on all sockets and seems to cause nginx to panic when declaired multiple times


The-Voidwalker:
Manual testing suggests this isn't a fix for the problem currently encountered on test131.

The-Voidwalker:
Deferred is also another one that can only be defined once. Problem is, it doesn't look to be default, and our configuration is too modular to know for certain whether or not another config block would have this enabled already. Not certain how to fix.

PR #3215, opened 2023-05-19
mariadb remove read-only on startup

The-Voidwalker:
mysql should automatically handle error recovery. There's no sense making it read-only especially if we don't have a procedure in place to validate the database after it's started up.

The only use case I can think of for having read-only on startup is for doing maintenance where the server is expected to restart, but we don't want any changes after the restart.


RhinosF1:
This was originally added in case mysql didn't start up cleanly (very possible in an uncontrolled shut down) so users were less likely to make a damaged database worse. This is much more important though if you have replication which we don't.

for maintenance, we can set it application level to reduce the risk.

paladox:
+1

PR #3281, opened 2023-07-09
check_reverse_dns: don't force mw-lb.miraheze.org

MacFan4000:
There is no technical reason that a custom domain couldn't point to some other miraheze.org subdomain. All possible subdomains will always exist and resolve to our servers. This uses regex to ensure that the CNAME is of the format <subdomain>.miraheze.org . The alert really comes into play if somebody has changed their DNS after the domain was setup for their wiki. And if they just changed what subdomain the record points to, I don't think it's worth chasing the user up about it. As for new domains, I always check the records beforehand (and anybody else handling SSL requests should too), and thus this check should not be relied on when adding SSL certs. We should be checking records before adding SSL, and only contacting users after if they changed it after, and for some reason the new configuration won't work. (non-miraheze.org CNAME, or improper A records, or somthing similar, or not pointed at all)


MacFan4000:
updated to disallow cp* and non-mediawiki subdomains (most) and non-mw subdomains not included, and other server names would already fail the RDNS check. (the left out non-mw subdomains are the ones not going though varnish)

RhinosF1:
It would be nice to see tests

modules/monitoring/files/check_reverse_dns.py
```diff
@@ -96,7 +96,11 @@ def check_records(hostname):
     except resolver.NoAnswer:
         cname = None
 
-    if cname == 'mw-lb.miraheze.org.':
+    if (
+        re.match("[A-Za-z]+.miraheze.org", cname) and
```
paladox, line 100: Subdomains can contain numbers.

modules/monitoring/files/check_reverse_dns.py
```diff
@@ -96,7 +96,11 @@ def check_records(hostname):
     except resolver.NoAnswer:
         cname = None
 
-    if cname == 'mw-lb.miraheze.org.':
+    if (
+        re.match(".+\.miraheze\.org", cname) and
+        not re.match("cp[1-9][0-9]*\.miraheze\.org", cname) and
```
Universal-Omega, line 101: I don't think we need this since hostnames use wikitide.net now

modules/monitoring/files/check_reverse_dns.py
```diff
@@ -96,7 +96,11 @@ def check_records(hostname):
...
+        not re.match("(webmail|phabricator|grafana|icinga|matomo|static|donate|reports)\.miraheze\.org", cname)
```
Universal-Omega, line 102: Remove webmail

Universal-Omega (changes requested):

MacFan4000:
> It would be nice to see tests

With the change a bit simpler now, I don’t think additional tests are needed.

RhinosF1:
> > It would be nice to see tests
> 
> With the change a bit simpler now, I don’t think additional tests are needed.

Tests are always good. This is one of the early targets for Python functions so I'll write them.

PR #3357, opened 2023-08-04
miraheze-backup: backup mediawiki-xml dumps to archive

paladox:
We don't want this to be backed up into our ovh backup as we already have sql backups. Instead back this up to archive.


RhinosF1 (changes requested):
Please don't use erb files for python. Should be a parameter / config file

PR #3474, opened 2023-10-23
switch test131 to mw 1.41

MacFan4000:


paladox:
For now we can do this. We might want to put it back on 1.40 to test php 8.

PR #3485, opened 2023-11-07
[BREAKING CHANGE] mwscript: make run.php no longer experimental

RhinosF1:
CHANGELOG

- Drops `--140`/`--use-runner` option (now default)
- adds `--no-runner` to switch to old behaviour
- added `syscheck` method
- `get_commands` can now return an int as well as CommandInfo. Result of get_commands can be passed to syscheck method which will pass int to sys.exit or return CommandInfo. Migration `X = get_commands(args)` becomes `X = syscheck(get_commands(args))`

Developer Notes:
- Added some extra tests
- `get_commands` no longer calls `sys.exit` at any point

### DIFF
--- a/modules/mediawiki/files/bin/mwscript.py
+++ b/modules/mediawiki/files/bin/mwscript.py
@@ -38,8 +38,7 @@ def get_commands(args: argparse.Namespace) -> CommandInfo:
         print('Not enough Arguments given.')
         sys.exit(2)
     script = args.script
-    if not script.endswith('.php'):
-        if args.norunphp:
+    if not script.endswith('.php') and args.norunphp:
             print('Error: Specifiy --use-runner or --140 to enable MaintenanceRunner')
             sys.exit(2)
     if not args.norunphp:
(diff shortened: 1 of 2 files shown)


modules/mediawiki/files/bin/mwscript.py
```diff
@@ -33,39 +39,35 @@ def get_commands(args: argparse.Namespace) -> CommandInfo:
...
-            print('Error: Specifiy --use-runner or --140 to enable MaintenanceRunner')
-            sys.exit(2)
-        if args.runner and not args.confirm:
-            print(f'WARNING: Please log usage of {longscripts}. Support for longscripts has not been added')
-            print('WARNING: Use of classes is not well tested. Please use with caution.')
-            if input("Type 'Y' to confirm (or any other key to stop - rerun without --140/--use-runner): ").upper() != 'Y':
-                sys.exit(2)
-    if args.runner:
+    if not script.endswith('.php') and args.norunphp:
+        print('Error: Specifiy --use-runner or --140 to enable MaintenanceRunner')
```
paladox, line 48: --use-runner and --140 seem to be gone in this change?

RhinosF1:
```suggestion
        print('Error: You can't use a class and specify --no-runner')
```

PR #3582, opened 2024-01-04
mariadb: remove SSL

Universal-Omega:


paladox:
Can we make this optional for now? I'll like to switch ssl off in the new cluster but keep it on in the old one.

PR #3584, opened 2024-01-05
Create rename-wiki.py

Reception123:


modules/mediawiki/files/cookbooks/rename-wiki.py
```diff
@@ -0,0 +1,53 @@
+import subprocess
+import argparse
+
+# Define the mapping of db clusters to db names
+db_clusters = {
```
RhinosF1, line 5: Make a Typed dict

modules/mediawiki/files/cookbooks/rename-wiki.py
```diff
@@ -0,0 +1,53 @@
...
+    'c1': 'db131',
+    'c2': 'db101',
+    'c3': 'db142',
+    'c4': 'db121',
+    'c5': 'db131',
+}
+
+
+def get_db_cluster(oldwiki_db):
```
RhinosF1, line 14: Should be typed. Oldwiki_db is a string. Returns a string.

modules/mediawiki/files/cookbooks/rename-wiki.py
```diff
@@ -0,0 +1,53 @@
...
+    command = f"salt-ssh -E 'db131*' cmd.run 'USE mhglobal; SELECT wiki_dbcluster FROM cw_wikis WHERE wiki_dbname = \"{oldwiki_db}\"'"
```
RhinosF1, line 15: Has this command been tested?

modules/mediawiki/files/cookbooks/rename-wiki.py
```diff
@@ -0,0 +1,53 @@
...
+    result = subprocess.run(command, shell=True, stdout=subprocess.PIPE, text=True)
+    cluster_name = result.stdout.strip()
+    return db_clusters.get(cluster_name)
+
+
+def execute_salt_command(cluster, command):
```
RhinosF1, line 21: Should be typed but I recommend splitting into generate_salt and execute_salt. That way you can confirm that your salt command is valid. You could possibly then use that in other places.

modules/mediawiki/files/cookbooks/rename-wiki.py
```diff
@@ -0,0 +1,53 @@
...
+    subprocess.run(salt_command, shell=True)
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Executes the commands needed to rename wikis')
+    parser.add_argument('--oldwiki', required=True, help='Old wiki database name')
+    parser.add_argument('--newwiki', required=True, help='New wiki database name')
+
+    args = parser.parse_args()
+
```
RhinosF1, line 32: Anything after here is logic

modules/mediawiki/files/cookbooks/rename-wiki.py
```diff
@@ -0,0 +1,53 @@
...
+    cluster_name = result.stdout.strip()
+    return db_clusters.get(cluster_name)
+
+
+def execute_salt_command(cluster, command):
+    salt_command = f"salt-ssh -E '{cluster}*' cmd.run '{command}'"
+    subprocess.run(salt_command, shell=True)
+
+
+def main():
```
RhinosF1, line 26: No type. Returns None. Maybe should return an exit code though.

RhinosF1 (changes requested):

modules/mediawiki/files/cookbooks/rename-wiki.py
Reception123: Not with salt but the actual command is the one I always use. I'll test them each individually of course.

RhinosF1: Check the full thing you've typed out. Is missing mysql -e

PR #3681, opened 2024-01-25
Install phorge171

paladox:


hieradata/hosts/phab171.yaml
MacFan4000, line 1: pretty sure it should be phorge171.yaml?

paladox: Yes I fixed that in a later commit, thanks

PR #3708, opened 2024-01-26
Use mem151/mem161 on new infra

Universal-Omega:


hieradata/common.yaml
```diff
@@ -183,9 +183,9 @@ swift::storage::devices:
 role::mediawiki::mcrouter::shards:
   miraheze:
     shard01:
-      host: [2a10:6740::6:510]
+      host: [2602:294:0:b13::105]
```
paladox, line 186: You can use the private ips (although I'd try it on the new mw first by changing the config manually to see if it'll work with the ipv4).

PR #3204, opened 2023-04-19
sshd: enforce use of PIN+Touch on FIDO SSH

RhinosF1:

PR #3725, opened 2024-01-31
Revert "sshd: enforce use of PIN+Touch on FIDO SSH"

redbluegreenhat:
Reverts miraheze/puppet#3204

My key malfunctions with this no matter what I do. Maybe we can do this when we all have Yubikey Bio keys, but not now.

### DIFF
--- a/modules/ssh/templates/sshd_config.erb
+++ b/modules/ssh/templates/sshd_config.erb
@@ -19,7 +19,6 @@ LogLevel VERBOSE
 
 # Authentication:
 LoginGraceTime 120
-PubkeyAuthOptions verify-required
 <% if @permit_root %>
 PermitRootLogin yes
 <% else %>


redbluegreenhat:
I should note you still need to physically have the key and need to 1. know the pin to generate the key to begin with and 2. need to physically have it when using it for it to work, so even without this it is still better than regular keys in terms of security.

redbluegreenhat:
I should add that I currently do not have access and need this to be merged.

If you don't like this change I'm open to replace my key with a regular ed25519.

RhinosF1:
Define malfunctions

redbluegreenhat:
Alright, so this is supposed to ask for me to touch the key AND input my FIDO PIN, right? Problem is my key doesn't do that. It only asks for my PIN when I'm initially creating the key, and only asks for me to touch the key when using it. I imagine the server isn't satisfied with that because I'm not also using my PIN, and so doesn't let me login. If you want logs ask UO.

RhinosF1:
Let's try and fix this within a task for future. Cc @Universal-Omega 

I think as long as you are sensible and immediately report a loss of the key via a security task and badgering the ops team (not publicly stating it until key revoked). It should be fine.

Ensure you practice good security rules. No writing this is key gets you into Miraheze.org next to it in case it gets stolen.

The risk of compromise via this route should be fairly minimal and in all honesty a physical key has a much smaller attack vector than a password unless you leave your laptop unlocked with the key plugged in while you go to the toilet in a cafe.

I do think there's a slight risk though and it's probably worth the risk being acknowledged at CTO level (cc @Universal-Omega again).

redbluegreenhat:
Good thing I don't have a laptop. Also, you need both the key and the encrypted identification on my computer at home, so that's a pretty tough achievement for anyone trying to enter to the servers via my SSH key.

RhinosF1:
Yeah my only suggestion is don't leave it plugged in (ideally keep it in a separate room away from external doors/windows so if you get broke into they can't take both easily. 

Its a pretty tough requirement and I should go to sleep though and honestly if someone breaks into an SRE's house after finding the address with the sole purpose of gaining access, they probably care too much and have enough resources we might as well let them.

PR #3739, opened 2024-02-05
cvtbot: fix WorkingDirectory and Environment

paladox:


Universal-Omega:
It has to be ran in net6.0 though, that is where the files are, there is nothing in Release.

PR #3814, opened 2024-03-31
update rename wiki script with "good version"

Reception123:


Universal-Omega (changes requested):
There is no c5 anymore since migration and we should not use /home/reception in puppetized script should use the current users puppet

RhinosF1:
> There is no c5 anymore since migration and we should not use /home/reception in puppetized script should use the current users puppet

#3817

modules/salt/files/bin/rename-wiki.py
```diff
@@ -9,6 +12,7 @@ class DbClusterMap(TypedDict):
     c2: str
     c3: str
     c4: str
+    c5: str
```
Universal-Omega, line 15: Still no c5 CC @RhinosF1

modules/salt/files/bin/rename-wiki.py
```diff
@@ -21,10 +25,10 @@ class DbClusterMap(TypedDict):
 
 
 def generate_salt_command(cluster: str, command: str) -> str:
-    return f'salt-ssh -E "{cluster}*" cmd.run "{command}"'
+    return f'salt-ssh -E "{cluster}" cmd.run "{command}"'
```
Universal-Omega, line 28: IIRC we don't need -E if we aren't using * but isn't * required if the full hostname is IE db151.wikitide.net? I can't remember if salt uses that or just db151 though

modules/salt/files/bin/rename-wiki.py
```diff
@@ -34,31 +38,40 @@ def execute_salt_command(salt_command: str, shell: bool = False, stdout: Optiona
...
-    execute_salt_command(salt_command=generate_salt_command(oldwiki_cluster, f"mysql -e 'CREATE DATABASE {newwiki_db}'"))
-    execute_salt_command(salt_command=generate_salt_command(oldwiki_cluster, f"mysql -e 'USE {newwiki_db}; SOURCE /home/$user/oldwikidb.sql'"))
+    execute_salt_command(salt_command=generate_salt_command(oldwiki_cluster, f'sudo -i mysqldump {oldwiki_db} > /home/{os.getlogin()}/{oldwiki_db}.sql'))
+    execute_salt_command(salt_command=generate_salt_command(oldwiki_cluster, f"sudo -i mysql -e 'CREATE DATABASE {newwiki_db}'"))
+    execute_salt_command(salt_command=generate_salt_command(oldwiki_cluster, f"sudo -i mysql -e 'USE {newwiki_db}; SOURCE /home/{os.getlogin()}/{oldwiki_db}.sql;'"))
 
     # Step 3: Execute MediaWiki rename script
-    execute_salt_command(salt_command=generate_salt_command('mwtask181', f'sudo -u www-data php /srv/mediawiki/1.41/extensions/CreateWiki/maintenance/renameWiki.php --wiki=loginwiki --rename {oldwiki_db} {newwiki_db} $user'))
+    execute_salt_command(salt_command=generate_salt_command('mwtask181', f'mwscript extensions/CreateWiki/renameWiki.php loginwiki --no-log --rename {oldwiki_db} {newwiki_db} {os.getlogin()}'))
```
Universal-Omega, line 72: I think mwtask181 should be some sort of constant so it is easy to change if task server is changed.

Universal-Omega (changes requested):

modules/salt/files/bin/rename-wiki.py
Reception123: Yeah that makes sense, how exactly would I do that here?

Universal-Omega: Similar to https://github.com/miraheze/puppet/blob/ac99c896596fa43eb01bb5000db17fa2f8506825/modules/mediawiki/files/bin/mwdeploy.py#L22

Universal-Omega (approved):

PR #3823, opened 2024-04-04
Add apt source for libssl1.1

Universal-Omega:


clarfonthey:
What does the server need that still requires libssl1.1? o.o

Universal-Omega:
> What does the server need that still requires libssl1.1? o.o

changeprop and eventgate bizarrely required for node to run it which I really disliked myself...

PR #3848, opened 2024-05-07
Add a --include-kernel option to do kernel upgrades as well

redbluegreenhat:


clarfonthey:
This script takes two arguments already but avoids the first? 🤔

But in this case, with the `--include-kernel` option, it's replacing the first?

redbluegreenhat:
> This script takes two arguments already but avoids the first? 🤔
> 
> But in this case, with the `--include-kernel` option, it's replacing the first?

Not sure what you mean here. I guess if you really want to mention it there's $0, the first argument, which is the name of the script, but this script never did anything with that.

This script never had any other arguments, so it's not replacing any options.

clarfonthey:
Oh, I completely misread the `print $2` in the awk command, that's entirely my bad.

Universal-Omega:
> > This script takes two arguments already but avoids the first? 🤔
> > But in this case, with the `--include-kernel` option, it's replacing the first?
> 
> Not sure what you mean here. I guess if you really want to mention it there's $0, the first argument, which is the name of the script, but this script never did anything with that.
> 
> This script never had any other arguments, so it's not replacing any options.

Just one minor correction here, $0 isn't always the actual script name, it can also be `-bash` if ran like `. script` because that is the name of the shell itself due to how it executes but $1 is still the first actual argument here always. Just a tiny technical correction.

Anyway this looks good to me.

PR #3852, opened 2024-05-10
Rename shell user

Universal-Omega:


clarfonthey:
Looks fine to me, assuming Puppet will actually handle renaming the users correctly.

This can be done later, but would it be possible to make the monitoring users draw from the user data in the Puppet config, so that you don't have to worry about emails etc. being in sync?

RhinosF1:
@clarfonthey: Puppet will not handle renaming users. It will have to be done manually.

RhinosF1:
> This can be done later, but would it be possible to make the monitoring users draw from the user data in the Puppet config, so that you don't have to worry about emails etc. being in sync?

if someone patches to user or monitoring module to work with that.

Universal-Omega:
> @clarfonthey: Puppet will not handle renaming users. It will have to be done manually.

Yeah that's why I didn't merge yet I'll handle rename manually later then merge this.

PR #3857, opened 2024-05-29
Mention python-functions repository on relevant scripts

clarfonthey:
After @RhinosF1 mentioned this to me, figured it would be good to add a comment saying these scripts will have a new home eventually.


redbluegreenhat:
CI for Python failed but this failure was not introduced by this PR. LGTM

PR #3876, opened 2024-07-04
mediawiki: prevent mwdeploy from upgrading extensions configured to checkout a specific commit

redbluegreenhat:
https://github.com/miraheze/mediawiki-repos supports configuring an extension to checkout a specific commit. However, if the extension is upgraded by mwdeploy, such as during `--upgrade-world`, then it will just pull the latest commit from the configured branch, which is not desired behavior. Puppet will restore the extension to the commit configured in mediawiki-repos, but by then the rsync has already happened.

This pull request aims to fix that by creating a file with the name of the extension on `/var/local/mwdeploy`, which if mwdeploy sees while upgrading an extension, will tell the script to not attempt to upgrade. This file is created automatically by Puppet, and removed automatically when the `commit` parameter for that extension is removed from mediawiki-repos

### DIFF
--- a/modules/mediawiki/files/bin/mwdeploy.py
+++ b/modules/mediawiki/files/bin/mwdeploy.py
@@ -441,7 +441,7 @@ def run_process(args: argparse.Namespace, version: str = '') -> list[int]:  # pr
                     if not os.path.exists(_get_staging_path(f'extensions/{extension}', version)):
                         print(f'{extension} does not exist for {version}. Skipping...')
                         continue
-                    if os.path.exists(_get_local_path(extension))
+                    if os.path.exists(_get_local_path(extension)):
                         print(f'Skipping upgrade of {extension} because it is configured in mediawiki-repos to checkout a specific commit')
                         continue
                     process = os.popen(_construct_git_pull(f'extensions/{extension}', submodules=True, quiet=False, version=version))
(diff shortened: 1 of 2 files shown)


Universal-Omega:
I think using git to check if a commit is checked out would be better than a file? Also we should also add tests here.

redbluegreenhat:
> I think using git to check if a commit is checked out would be better than a file? Also we should also add tests here.

Puppet won't touch the extension if the commit parameter is removed in mediawiki-repos as long as the extension exists. So I think there's no way to differentiate then between an extension configured to checkout a specific commit to one that used to be, but is no longer configured to do so without doing this. I'm probably wrong tho

I'll work on the tests

Universal-Omega:
> > I think using git to check if a commit is checked out would be better than a file? Also we should also add tests here.
> 
> Puppet won't touch the extension if the commit parameter is removed in mediawiki-repos as long as the extension exists. So I think there's no way to differentiate then between an extension configured to checkout a specific commit to one that used to be, but is no longer configured to do so without doing this. I'm probably wrong tho
> 
> I'll work on the tests

I don't know what you mean? If you use git to check it doesn't rely on puppet it uses the current value and check for something like a "Detached HEAD" state.

Universal-Omega:
IE use something like the command `git symbolic-ref -q HEAD` and return True or False based on exit code in a new function, `is_detached_head`, use that function to filter our extensions & skins from `get_valid_extensions` and `get_valid_skins`

PR #3885, opened 2024-07-17
mwdeploy.py: add support for local patches

The-Voidwalker:
still a bit WIP


Universal-Omega (approved):

modules/mediawiki/manifests/deploy.pp
```diff
@@ -70,6 +70,51 @@
...
+        mode    => '0755',
+        require => File['/srv/mediawiki-staging'],
+    }
+
+    file { '/srv/mediawiki-staging/patches/public.json':
+        ensure  => 'file',
+        owner   => 'www-data',
+        group   => 'www-data',
+        mode    => '0644',
+        source  => 'puppet:///modules/mediawiki/patches/public.json',
```
Universal-Omega, line 86: I wonder if it would be better to do these in mediawiki-repos so mediawiki-admins can push without ops review?

The-Voidwalker: That's a pretty good idea, I'll take a look at that soon.

Universal-Omega: * @The-Voidwalker #3940 / miraheze/mediawiki-repos#35

pixDeVl: > * @The-Voidwalker [Use mediawiki-repos for public patches #3940](https://github.com/miraheze/puppet/pull/3940) / [Add patches directory mediawiki-repos#35](https://github.com/miraheze/mediawiki-repos/pull/35)

@Universal-Omega Is there anything else blocking this PR that needs to be added for merging and deployment besides #3940 ?

modules/mediawiki/files/bin/mwdeploy.py
```diff
@@ -332,6 +341,35 @@ def _construct_reset_mediawiki_run_puppet() -> str:
...
 
+def _construct_git_apply(repo: str, patchfile: str, version: str = '', check: bool = False) -> str:
+    extopt = ' --check' if check else ' --index'
+    return f'sudo -u {DEPLOYUSER} git -C {_get_staging_path(repo, version)} apply{extopt} {patchfile}'
+
+
+def _apply_patches(repo: str, version: str = '') -> list[int]:
+    exitcodes = []
+    staging_path = _get_staging_path(repo, version)
+    to_apply = [patch for patch in patches if staging_path.endswith(patch.path)]
```
BlankEclair, line 352: Do we have to do it in a list comprehension instead of the for loop? And did you mean startswith()?

And also, each `patch` is a dictionary

And also also, according to https://github.com/miraheze/mediawiki-repos/blob/9c0ef97ea390c3d36bc52f39bef2db8c91147777/patches/public/README.md, the path is relative, and there's a `version` key to filter for things

modules/mediawiki/files/bin/mwdeploy.py
```diff
@@ -332,6 +341,35 @@ def _construct_reset_mediawiki_run_puppet() -> str:
...
+    for patch in to_apply:
+        pubpriv = 'public' if patch.public else 'private'
+        patchfile = f'/srv/mediawiki-staging/patches/{pubpriv}/{patch.file}'
```
BlankEclair, line 355:
```suggestion
        visibility = 'public' if patch.public else 'private'
        patchfile = f'/srv/mediawiki-staging/visibility/{pubpriv}/{patch.file}'
```

modules/mediawiki/files/bin/mwdeploy.py
Universal-Omega: endswith is correct isn't it? Since it's relative staging path would end in extensions/CreateWiki for example if extensions/CreateWiki is the patch path since staging path is `f'/srv/mediawiki-staging/{version}/{repo}'`

RhinosF1:
@The-Voidwalker: What's next here?

The-Voidwalker:
> @The-Voidwalker: What's next here?

I'm a bit out of the loop myself (focusing on some career stuff lately), but I think testing on beta would be a good place to start. I don't think I can work on it immediately, but if anyone else wants to take ownership, please do (and leave a comment so the rest of us can know).

Universal-Omega (approved):
I have finished this up to a point where it at least works as expected, at least seems to for me when testing on beta. Future work can be done on this later though. Thanks for doing this! I am merging so that we can use this for patches now.

PR #3886, opened 2024-07-19
Revert "temporarily disable error reporting for E_DEPRECATED"

The-Voidwalker:
This reverts commit fe0398388a588fbfc1d61a418bd447467406d744.


The-Voidwalker:
Should wait until we can confirm this is fixed.

PR #3893, opened 2024-08-02
Pywikibot: Add user agent description

redbluegreenhat:


modules/irc/templates/pywikibot/user-config.py
```diff
@@ -2,3 +2,4 @@
 mylang = 'meta'
 usernames['wikitide']['meta'] = 'BeeBot'
 authenticate['*.miraheze.org'] = ('<%= @consumer_token %>', '<%= @consumer_secret %>', '<%= @access_token %>', '<%= @access_secret %>')
+user_agent_description = 'https://meta.miraheze.org/wiki/Tech:Pywikibot; tech@miraheze.org'
```
Universal-Omega, line 5: I think we can use @wikitide.org as foundation primary email. We don't need to use an alias here.

PR #3917, opened 2024-09-08
T12559: Workaround resources not loading for non-Cloudflare domains

BlankEclair:


RhinosF1 (approved):
Hacky but this is only going to last a couple of weeks at most

redbluegreenhat:
I've created an internal CA that is able to issue certificates that NGINX will trust (once #3923 is merged) for client authentication, instead of this a client certificate will be issued for the cache proxies that they can use for communicating with mw*

PR #3918, opened 2024-09-08
Use better Firejail wrappers

BlankEclair:
The python scripts:
* Hide the return code
* Does a fork() + exec() instead of just an exec()
* Could've easily been a shell script

Therefore... we do it in a shell script that fixes the above!

## Summary by CodeRabbit

- **New Features**
	- Introduced multiple shell scripts to execute various commands (convert, espeak, ghostscript, lame, rsvg-convert) within a secure Firejail sandbox environment, enhancing security for these processes.
  
- **Bug Fixes**
	- Updated source paths in the manifest to reflect the transition from Python scripts to the new shell scripts, ensuring proper execution.


RhinosF1:
@coderabbitai review

PR #4027, opened 2024-10-14
monitoring: only apply ssl service if the domain doesn't end in .miraheze.org

MacFan4000:
This reduces the number of redundant checks for miraheze.org, which already has a check defined towards the beginning of this file.


paladox:
We need the ssl cert expiry monitoring. But also events are disabled via disable_event.

MacFan4000:
> We need the ssl cert expiry monitoring. But also events are disabled via disable_event.

As mentioned above the miraheze.org cert is already monitored by a separate entry in the config. There is zero reason to have it monitored many times extra.

RhinosF1 (approved):
we should only be monitoring things we can actually do something about, there are better ways of monitoring cf domains

PR #4075, opened 2024-11-12
T12829: Use a different user for pywikibot

redbluegreenhat:
https://issue-tracker.miraheze.org/T12829


Universal-Omega:
> @Universal-Omega should I define the user in irc::pywikibot or in the role module? A search of the puppet repo reveals all users are defined in role, except www-data, which is created already by debian and is only on puppet to add an SSH key to it on mediawiki.

in role is best

PR #4102, opened 2024-11-22
mediawiki::deploy: require manual deployments

paladox:


RhinosF1 (changes requested):
Probably should absent the timers first before removal

RhinosF1 (approved):

PR #4115, opened 2024-12-10
T12621 - restrict what is considered a php file

paladox:
Re-do of #4112 by @The-Voidwalker. The correct syntax is provided by @BlankEclair.

### DIFF
--- a/modules/mediawiki/templates/mediawiki-includes.conf.erb
+++ b/modules/mediawiki/templates/mediawiki-includes.conf.erb
@@ -17,10 +17,10 @@ location /w/rest.php/ {
 	try_files $uri $uri/ /w/rest.php?$args;
 }
 
-location ~ \.php {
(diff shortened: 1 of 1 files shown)


modules/mediawiki/templates/mediawiki-includes.conf.erb
```diff
@@ -17,10 +17,10 @@ location /w/rest.php/ {
 	try_files $uri $uri/ /w/rest.php?$args;
 }
 
-location ~ \.php {
+location ~ ^/(w/)?\w+\.php {
```
BlankEclair, line 20: To copy my former review: What scripts are we loading from the document root?

BlankEclair: Plenty actually: https://issue-tracker.miraheze.org/P532

BlankEclair (approved):
Fingers crossed...

PR #4116, opened 2024-12-10
touch.php and favicon.php: Follow redirects

paladox:


BlankEclair (approved):
While the docs do say "this should only be used when the target URL is trusted, to avoid attacks on intranet services accessible by HTTP", we already have plenty of SSRFs, so I don't think it matters much?

Don't forget about favicon.php too

BlankEclair (commented):
Instead of following redirects when it's static.{miraheze.org,wikitide.net}, we can instead parse the URL (ideally with UrlUtils), check if it's static.miraheze.org, and if so, recreate the URL but with the domain as static.wikitide.net

Essentially, handle that specific redirect ourselves

modules/mediawiki/files/touch.php
```diff
@@ -26,7 +26,16 @@ function streamAppleTouch() {
...
 
 	$url = wfExpandUrl( $touch, PROTO_CANONICAL );
-	$client = MediaWikiServices::getInstance()
+
+	$mediaWikiServices = MediaWikiServices::getInstance();
+	$urlUtils = $mediaWikiServices->getUrlUtils();
+	$parsedBaseUrl = $urlUtils->parse( $url );
+
+	if ( $parsedBaseUrl['host'] === 'static.miraheze.org'  ) {
+		$url = str_replace( 'static.miraheze.org', 'static.wikitide.net', $url );
```
BlankEclair, line 35: paladox no...

https://doc.wikimedia.org/mediawiki-core/master/php/classMediaWiki_1_1Utils_1_1UrlUtils.html#a74f7394421912d67d6aa343060c4fb7a

paladox: Thanks, done.

PR #4119, opened 2024-12-12
switch cp* default ssl to CF origin cert

MacFan4000:
This is the final thing that still uses the now expired section cert. This also removes the cert definition for the old cert from the SSL module.

### DIFF
--- a/modules/varnish/templates/mediawiki.conf.erb
+++ b/modules/varnish/templates/mediawiki.conf.erb
@@ -98,8 +98,8 @@ server {
 	server_name miraheze.org *.miraheze.org;
 	root /var/www/html;
 
-	ssl_certificate /etc/ssl/localcerts/wildcard.miraheze.org-2020-2.crt;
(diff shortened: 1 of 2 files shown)


modules/varnish/templates/mediawiki.conf.erb
```diff
@@ -98,8 +98,8 @@ server {
 	server_name miraheze.org *.miraheze.org;
 	root /var/www/html;
 
-	ssl_certificate /etc/ssl/localcerts/wildcard.miraheze.org-2020-2.crt;
```
paladox, line 101: This will break custom domains using varnish. Cloudflare cert can only be used with Cloudflare.

RhinosF1: No it won't - they'll use their block - it's an expired cert anyway, we don't need a default block at all tbh though. We should only be serving domains which have explicit configuration. We could remove the miraheze.org & default blocks altogether.

PR #4134, opened 2024-12-23
varnish: Allow caching mobile page

paladox:


paladox:
I'm not sure why this was changed in https://github.com/miraheze/puppet/commit/ec1f216641b26e7b36eaef67e83a91fe71a2cc25#diff-f9b990bb03dfc46d5ffe4daec99c0fe8f7fe5fc0998ad48c3119560125bbfb6fL418

but the hashing of it is what seperates the caches. So e.g. if X-Subdomain is empty and then if it's set with M it'll save two versions
the M is the mobile one 

https://www.mediawiki.org/wiki/Extension:MobileFrontend/Configuring_browser_auto-detection seems to imply wgCookieDomain needs to be set

PR #4177, opened 2025-02-01
base: remove pdns-recursor package if dns hiera setting is set

paladox:


RhinosF1:
NB: actionable from https://meta.miraheze.org/wiki/Tech:Incidents/2025-02-01-dns

pdns-recursor isn't compatible with GDNSD so both shouldn't be installed on the same server

PR #4187, opened 2025-02-13
mediawiki: hash based on header for jobrunner haproxy config

paladox:


freephile:
Depending on test results, this may resolve https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5929

PR #4212, opened 2025-02-27
fix backups log location for grafana

MacFan4000:


modules/role/manifests/grafana.pp
```diff
@@ -45,7 +45,7 @@
     }
 
     monitoring::nrpe { 'Backups Grafana':
-        command  => '/usr/lib/nagios/plugins/check_file_age -w 1382400 -c 1468800 -f /var/log/grafana-backup.log',
+        command  => '/usr/lib/nagios/plugins/check_file_age -w 1382400 -c 1468800 -f /var/log/grafana-backup/grafana-backup/grafana-backup.log',
```
Universal-Omega, line 48: This is wrong to many grafana-backup

Universal-Omega:
```suggestion
        command  => '/usr/lib/nagios/plugins/check_file_age -w 1382400 -c 1468800 -f /var/log/grafana-backup/grafana-backup.log',
```

PR #4261, opened 2025-03-14
Run shellcheck

BlankEclair:
This commit has various shell scripts fixed after running shellcheck. However:
* I did not try to handle *.erb files
* I didn't even test these (maybe production will blow up? who knows!)
* I didn't bother with a few warnings lol


The-Voidwalker (changes requested):
We don't want any accidental variable expansion here, it'll break the script.

PR #4274, opened 2025-03-23
T13404: Do not redirect for /w/index.php?action=render

BlankEclair:
Fixes T13404


AgentIsai:
There're [quite a few action parameters to index.php](https://www.mediawiki.org/wiki/Manual:Parameters_to_index.php#Actions). Are we sure we want to stop redirecting all of them?

Universal-Omega:
We want some actions like edit to redirect though to normalize URLs otherwise we may as well remove the whole thing as that was the entire point of it.

BlankEclair:
Fair

Universal-Omega (approved):

PR #4286, opened 2025-04-03
Create sql/eval/shell.sh

RhinosF1:
## Summary by CodeRabbit

- **New Features**
	- Introduced three new command-line utilities to safely execute tasks without confirmation prompts and controlled logging.
- **Chores**
	- Enhanced deployment automation to install these utilities with proper executable permissions.

### DIFF
--- a/modules/mediawiki/files/bin/eval.sh
+++ b/modules/mediawiki/files/bin/eval.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+mwscript eval "$1" --no-log --confirm
--- a/modules/mediawiki/files/bin/eval.sh
+++ b/modules/mediawiki/files/bin/eval.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-mwscript eval "$1" --no-log --confirm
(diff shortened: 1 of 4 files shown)


RhinosF1:
@coderabbitai review

BlankEclair (commented):
What if we used this stub instead?
```shell
#!/bin/sh
wiki="$1"
shift
if [[ -z "$wiki" ]]; then
    echo "Usage: $0 <wiki> [extra arguments]..." >&2
    exit 1
fi

exec mwscript --no-log --confirm sql "$wiki" -- "$@"
```

RhinosF1:
> What if we used this stub instead?
> 
> ```shell
> #!/bin/sh
> wiki="$1"
> shift
> if [[ -z "$wiki" ]]; then
>     echo "Usage: $0 <wiki> [extra arguments]..." >&2
>     exit 1
> fi
> 
> exec mwscript --no-log --confirm sql "$wiki" -- "$@"
> ```

mwscript already validates the argument and we don't wanna allow passing an SQL file as you should log that

PR #4336, opened 2025-04-26
ssl: switch off automatic cert renewal for custom domains

RhinosF1:
No certificates are expected to require automatic renewal between now and May 16th when our Hosted DNS service is due to close. Therefore, these can be let to expire as all certs by then will be managed by Cloudflare.


MacFan4000 (approved):

Universal-Omega (approved):

RhinosF1:
@coderabbitai review

PR #4337, opened 2025-04-27
Replace LetsEncrypt.crt with ISRG_Root_X1.pem

RhinosF1:
I ran 
> [rhinos@test151:/etc/ssl/certs]$ diff LetsEncrypt.crt ISRG_Root_X1.pem
> [rhinos@test151:/etc/ssl/certs]$ 

and it's the same file


RhinosF1:
@coderabbitai review

PR #4402, opened 2025-06-17
mwdeploy: update gitinfo on --upgrade-{extensions,skins}

RhinosF1:


Universal-Omega:
Done as part of #4813

PR #4416, opened 2025-06-20
wikitide-backup: improve sql dumps to use mydumper and cleanup

Universal-Omega:


Universal-Omega:
This causes heavier memory usage

PR #4465, opened 2025-08-04
add SSL checks based on the cloudflare_domains file

MacFan4000:


Universal-Omega:
I don't really think this is needed. If others disagree this can be reopened.

PR #4516, opened 2025-09-09
mwscript: use metawikibeta if on beta; switch to meta wiki on prod (as loginwiki is deprecated)

MacFan4000:
This allows us to run for specific extensions on beta. Previously did not work, due to us hardcoding the use of prod loginwiki.


MacFan4000:
remaining build failure is unrelated to this PR

PR #4837, opened 2026-03-31
T15160: Set slave_connections_needed_for_purge to 0

SomeMWDev:
Required to unbreak binlog purging - see https://jira.mariadb.org/browse/MDEV-38849


SomeMWDev:
We need to remove this once we have replication, but we don't right now AFAIK, so this is a temporary solution

PR #4841, opened 2026-04-22
T15231: experimental robots.php adjustment for non-English support

lihaohong6:
No idea if this works. 

Works together with https://github.com/miraheze/mw-config/pull/6381 which adds the `REPLACEME` placeholder.


lihaohong6:
robots.php got moved to mw-config. This PR is no longer applicable.

PR #4846, opened 2026-05-09
T15193: Add profile for xvfb-run

pskyechology:


The-Voidwalker:
T15193 was resolved elsewhere, don't think this is still needed.

PR #4851, opened 2026-05-11
Beta: switch default version to 1.46

YTFGolf:


YTFGolf:
Apparently not to be done just yet

PR #4852, opened 2026-05-14
add global rename bot T13545

Reception123:
Approved in https://meta.miraheze.org/wiki/Requests_for_Comment/Partially_automate_global_renames

Note: This patch is mostly written by LLMs with adjustments. It has been tested on Beta and a limited test on Meta was also successful.

### DIFF
--- a/manifests/site.pp
+++ b/manifests/site.pp
@@ -5,6 +5,7 @@ node /^bast1[68]1\.wikitide\.net$/ {
 }
 
 node 'bots171.fsslc.wtnet' {
+    include mediawiki::rename_bot
     role(irc)
 }
 
--- a/modules/mediawiki/files/bots/policy_checker.py
+++ b/modules/mediawiki/files/bots/policy_checker.py
@@ -1,0 +1,246 @@
+"""
+Checks proposed usernames against Miraheze's Username Policy and
+MediaWiki:Global_title_blacklist.
+https://meta.miraheze.org/wiki/Username_Policy
+https://meta.miraheze.org/wiki/MediaWiki:Global_title_blacklist
+"""
+
+import re
+import unicodedata
+from typing import Optional
+
+# Roles/positions that would create a false impression of authority.
+AUTHORITY_TERMS = {
+    "admin", "administrator", "sysop",
+    "bureaucrat", "interface-admin", "interfaceadmin",
+    "checkuser", "oversight", "suppress",
+    "steward", "global-admin", "globaladmin",
+    "miraheze-staff", "mirahezestaff", "wikitide-staff", "wikitidefoundation",
+    "trustandsafety", "trust-and-safety",
+    "staff", "moderator", "mod", "cvt", "global-moderator",
+    "director", "official", "founder", "owner", "operator",
+    "developer", "support", "helpdesk",
+    "sysadmin", "superadmin", "root", "sudo",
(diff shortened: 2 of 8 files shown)


dmehus:
I love the idea of a renaming bot taking care of most global renames. Love that it even went to an RfC, which passed unanimously.

My only suggestion, @Reception123, would you be open to making a friendly amendment with the redirect suppression flag enabled when it performs the renames?

manifests/site.pp
```diff
@@ -5,6 +5,7 @@
 }
 
 node 'bots171.fsslc.wtnet' {
+    include mediawiki::rename_bot
```
Universal-Omega, line 8: This should really be refactored, into a new bots module along with IRC and relay bots, we shouldn't be loading stuff from MediaWiki module on bots171. We could put it on mwtask though similar to techdocs script. I can work on refactoring the IRC module in another PR though.

modules/mediawiki/files/bots/policy_checker.py
```diff
@@ -0,0 +1,246 @@
...
+
+BOT_TERMS = {
+    "bot", "script", "robot", "crawler", "spider", "scraper",
+    "automated", "autobot", "botaccount",
+}
+
+# All patterns translated directly from MediaWiki:Global_title_blacklist.
+# The blacklist is case-insensitive by default; all patterns use re.IGNORECASE.
+# re.search() is used (equivalent to the blacklist's leading/trailing .*).
+_BLACKLIST_PATTERNS = [
```
Universal-Omega, line 35: Name this something else to align in the change in naming scheme I think. Maybe `_BLOCKLIST_PATTERNS` or something.

modules/mediawiki/manifests/global_rename_bot.pp
```diff
@@ -0,0 +1,63 @@
...
+            source  => "puppet:///modules/mediawiki/bots/${f}",
+            require => File[$bot_dir],
+        }
+    }
+
+    package { ['python3-requests', 'python3-bs4']:
+        ensure => present,
+    }
+
+    cron { 'global_rename_bot':
```
Universal-Omega, line 49: Use systemd timer not cron.

modules/mediawiki/manifests/global_rename_bot.pp
```diff
@@ -0,0 +1,63 @@
...
+            ensure  => file,
+            owner   => $user,
+            group   => $user,
+            mode    => '0550',
+            source  => "puppet:///modules/mediawiki/bots/${f}",
+            require => File[$bot_dir],
+        }
+    }
+
+    package { ['python3-requests', 'python3-bs4']:
```
Universal-Omega, line 45: `stdlib::ensure_packages(['python3-bs4', 'python3-requests'])`

modules/role/manifests/mediawiki/bots.pp
```diff
@@ -0,0 +1,8 @@
+# role: mediawiki::bots — not used standalone; bots171 runs role::irc which includes this
+class role::mediawiki::bots {
```
Universal-Omega, line 2: This is not used at all and can be deleted.

Universal-Omega (changes requested):
Not sure if this is still needed or not since the bot does seem to be running already, but if so here is an initial review. Also the python CI checks need to be fixed to pass.

PR #4859, opened 2026-06-01
move all systemd units to /etc/systemd/system

The-Voidwalker:
`/lib/systemd/` is for vendor unit files and will be overwritten during updates. `/etc/systemd` is the preferred location for created units, as well as overrides. If a unit is defined in both locations `/etc/systemd` takes priority.

Fixes T15466

### DIFF
--- a/modules/systemd/manifests/init.pp
+++ b/modules/systemd/manifests/init.pp
@@ -5,7 +5,7 @@
 #
 class systemd {
     # Directories for base units and overrides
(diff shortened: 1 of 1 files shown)


The-Voidwalker:
Not willing to merge this myself just yet as it will trigger a restart of _all_ services.

Universal-Omega:
Can this be made a hiera, so that we can do a rolling deploy of it?

PR #4899, opened 2026-08-09
mwdeploy: add new feature and improvements

The-Voidwalker:
Improve reset-world and patch matching:
Fixes reset world to apply patches after the reset operations instead of
before. Additionally, for the purposes of following install steps,
supplying --reset-world now behaves like supplying --world.

Finally, for checking if a patch applies to a repo, we can assume it
does if the patch path is equal to the repo name and is a MediaWiki
version number. This simplifies our ability to apply patches to
MediaWiki core as we currently otherwise have weird setups we need to
follow in order to make it work.

add --apply-patches option:
This option provides the ability to apply patches without needing to
perform other updates.


Universal-Omega (approved):
lgtm

PR #4902, opened 2026-08-09
T15798: add suwd.sh to easily run commands as www-data

pskyechology:
Resolves T15798


RhinosF1 (approved):
I like suwd as a shorthand :)

PR #4936, opened 2026-08-25
mediawiki_beta: use internal network

Universal-Omega:


Universal-Omega:
Done in #4937 instead

PR #5021, opened 2026-09-02
backup: Switch backup script to use ovh object storage

paladox:


Universal-Omega (approved):

modules/backup/templates/wikitide-backup.py.epp
Universal-Omega, line 1: Meant to put in initial review, but there is room for improvements of this later. Would like to just get this done for now and clean it up later. Primarily better conforming to flake8 standards and use parameters to pass template args so the python script doesn't need to be an epp template at all.

